PR #8, opened 2020-06-21
Update README.md

adtrevor:
Update with correctly link to WASI, for some browsers at least the current URL cannot be opened.


MaxDesiatov:
Thank you @adtrevor!

PR #18, opened 2020-07-02
Add --destination option to the `carton dev` command

ratranqu:
- this add an optional flag to the `dev` command from `carton` to enable passing a `destination.json` file to the `swift build` call
- this requires that the toolchain referenced in the `destination.json` file is consistent with the one used by the `carton` process.
- it enables the importing of the `Foundation` module amongst others when running `carton dev`

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,40 @@ for WebAssembly and served when you start `carton dev` in the directory where `P
 [OpenCombine](https://github.com/broadwaylamb/OpenCombine), and supports both macOS and Linux. (Many
 thanks to everyone supporting and maintaining those projects!)
 
+### Providing a destination file to `carton dev`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -65,6 +65,40 @@ for WebAssembly and served when you start `carton dev` in the directory where `P
 [OpenCombine](https://github.com/broadwaylamb/OpenCombine), and supports both macOS and Linux. (Many
 thanks to everyone supporting and maintaining those projects!)
 
+### Providing a destination file to `carton dev`
+The `carton dev` command can be passed an optional destination.json file to the `swift build` command it calls. Currently, this is required to be able to use `Foundation` in your code.
```
MaxDesiatov, line 69: Just a minor formatting suggestion 🙂

```suggestion
The `carton dev` command can be passed an optional `destination.json` file to the `swift build` command it calls. Currently, this is required to be able to use `Foundation` in your code.
```

Issue #15, opened 2020-07-01
More commands for `carton sdk`


RayZhao1998:
I'v seen there's only one command `install` for `carton sdk`.

If we have the situation of consisting multiple versions of sdk, I think we need to implement more commands for `carton sdk`. 

- `carton sdk list`: Show all installed sdks.
- `carton sdk use`: Use a specific version of sdk you want.

MaxDesiatov:
Hi @RayZhao1998, it is absolutely reasonable, thank you for creating this issue! One point that needs to be resolved is how `carton sdk use` would work. Would it record that version in the `.swift-version` file? A lot of toolchain managers allow specifying a "global" and a "local" toolchain, so maybe instead of `carton sdk use` it would be `carton sdk global` and `carton sdk local`? IDK, just spitballing here, mostly looking for a consensus on what the best user experience would be 🙂

RayZhao1998:
> A lot of toolchain managers allow specifying a "global" and a "local" toolchain, so maybe instead of `carton sdk use` it would be `carton sdk global` and `carton sdk local`?

It make sense! we can use a flag such as `-g` representing "global", just like `npm install -g` or just use "global" in the command like `yarn global add`.

MaxDesiatov:
Yes, something like that, I personally prefer `global` and `local` subcommands [as `swiftenv` already uses them](https://swiftenv.fuller.li/en/latest/commands.html#global). And since `carton` already picks up existing `swiftenv` installation, maybe `carton` should follow their convention to make it easier for developer to switch.

RayZhao1998:
TODO:
- [x] `carton sdk versions`: show the list of installed sdks #21
- [ ] `carton sdk local`: show the local sdk version
- [ ] `carton sdk local [version]`: set the local sdk version
- [ ] `carton sdk global`: show the global sdk version
- [ ] `carton sdk global [version]`: set the global sdk version

### `carton sdk versions`

This command will look up sdk installed in `~/.carton/sdk` and `~/.swiftenv/versions`. It'll display which version is set as local and global.

### `carton sdk local`

In a project, the sdk version is saved in `.swift-env`. So we just need to look up the current local sdk version saved in `.swift-env`. Also when we want to set the local version of sdk, we just need to save the version to `.swift-env`.

### `carton sdk global`

`swiftenv` save the global version of swift in `~/.swiftenv/version`, maybe we can save our global version in `~/.carton/swift-version` too.

PR #21, opened 2020-07-03
Support `carton sdk versions`

RayZhao1998:
Move previous description to #15 .

This PR implements `carton sdk versions` to list all installed sdks.

### DIFF
--- a/Sources/carton/Toolchain/ToolchainManagement.swift
+++ b/Sources/carton/Toolchain/ToolchainManagement.swift
@@ -61,6 +61,14 @@ enum ToolchainError: Error, CustomStringConvertible {
 }
 
 extension FileSystem {
+  private var swiftenvVersionsPath: AbsolutePath {
+    return homeDirectory.appending(components: ".swiftenv", "versions")
+  }
+  
+  private var cartonSDKPath: AbsolutePath {
+    return homeDirectory.appending(components: ".carton", "sdk")
+  }
+  
   private func inferSwiftVersion(
     from versionSpec: String? = nil,
     _ terminal: TerminalController
@@ -119,11 +127,11 @@ extension FileSystem {
       return swiftPath.pathString
     }
 
-    if let path = try checkAndLog(homeDirectory.appending(components: ".swiftenv", "versions")) {
+    if let path = try checkAndLog(swiftenvVersionsPath) {
       return path
     }
 
-    let sdkPath = homeDirectory.appending(components: ".carton", "sdk")
+    let sdkPath = cartonSDKPath
     if let path = try checkAndLog(sdkPath) {
       return path
     }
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
I'd suggest saving the global version to `~/.carton/swift-version` instead of `~/.carton/version`. Just in case we'd like to use `~/.carton/version` later for staring the version of `carton` itself for some reason, while the name of `~/.carton/swift-version` seems to be quite unambiguous.

Sources/carton/Toolchain/ToolchainManagement.swift
```diff
@@ -282,4 +282,18 @@ extension FileSystem {
 
     return AbsolutePath(binPath)
   }
+  
+  func fetchAllSwiftVersion() throws -> [String] {
+    var versions: [String] = []
+    // sdks installed by carton
+    let cartonSdkPath = homeDirectory.appending(components: ".carton", "sdk")
```
MaxDesiatov, line 289: **suggestion:** could you separate out `homeDirectory.appending(components: ".carton", "sdk")` into a computed property within this same extension? Computing this path already happens in this same file on line 126, would be great to avoid duplication. Also, a more suitable name would be `cartonSDKPath` to keep the capitalization consistent.

Sources/carton/Toolchain/ToolchainManagement.swift
```diff
@@ -282,4 +282,18 @@ extension FileSystem {
...
+    let cartonSdks = try getDirectoryContents(cartonSdkPath)
+    versions.append(contentsOf: cartonSdks)
+    // sdks installed by swiftenv
+    let swiftenvVersionPath = homeDirectory.appending(components: ".swiftenv", "versions")
```
MaxDesiatov, line 293: **suggestion:** same for this path, it's already computed on line 122, would be great to generalize it.

Sources/carton/Toolchain/ToolchainManagement.swift
```diff
@@ -282,4 +282,18 @@ extension FileSystem {
 
     return AbsolutePath(binPath)
   }
+  
+  func fetchAllSwiftVersion() throws -> [String] {
```
MaxDesiatov, line 286:
```suggestion
  func fetchAllSwiftVersions() throws -> [String] {
```

Sources/carton/Toolchain/ToolchainManagement.swift
```diff
@@ -282,4 +282,18 @@ extension FileSystem {
...
+    // sdks installed by carton
+    let cartonSdkPath = homeDirectory.appending(components: ".carton", "sdk")
+    let cartonSdks = try getDirectoryContents(cartonSdkPath)
+    versions.append(contentsOf: cartonSdks)
+    // sdks installed by swiftenv
+    let swiftenvVersionPath = homeDirectory.appending(components: ".swiftenv", "versions")
+    let
+      swiftenvVersions = try getDirectoryContents(swiftenvVersionPath)
+    versions.append(contentsOf: swiftenvVersions)
+    return versions
```
MaxDesiatov, line 297: After you've separated out both of those path into computed properties, the function would fit on a single linee 🙂

```swift
func fetchAllSwiftVersions() throws -> [String] {
  try getDirectoryContents(cartonSDKPath) + getDirectoryContents(swiftenvVersionPath)
}
```

RayZhao1998:
Fixed! Please review again.

PR #33, opened 2020-07-08
Bump OpenCombine version to 0.10.0

broadwaylamb:
The new version contains important bug fixes that prevent memory leaks.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -10,7 +10,7 @@ let package = Package(
     .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.1.1"),
     .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.2.0")),
     .package(url: "https://github.com/apple/swift-tools-support-core.git", .upToNextMinor(from: "0.1.3")),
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
I think I really should use `from` instead of `upToNextMinor` on it, I don't think we'd expect the API to break even up to 1.0.

broadwaylamb:
> I think I really should use `from` instead of `upToNextMinor` on it, I don't think we'd expect the API to break even up to 1.0.

I'm not sure. What concerns me is the `OCombine` namespaces. There are some difficulties there, for example, if both class and its subclass need to have those namespaces (e. g. `NSObject.OCombine` for KVO publishing, and `URLSession.OCombine` for URLSession publisher). We may need to rethink the namespacing mechanism, and it can be source-breaking.

If you don't use the classes that may be affected my that issue, then it's okay to use `from`.

MaxDesiatov:
We do have `Package.resolved` in the repository, which is updated by the bot through PRs, so I think the risk of someone installing a version of OpenCombine that breaks the build is very minimal (they really would have to remove `Package.resolved` manually on their side as far as I understand). And if the bot creates a PR with a breaking update, better for me then as I'll be notified of the breakage quickly.

PR #40, opened 2020-07-13
Implement `carton sdk local` subcommand

RayZhao1998:
Support `carton sdk local` subcommand. #15

### DIFF
--- a/Sources/SwiftToolchain/ToolchainManagement.swift
+++ b/Sources/SwiftToolchain/ToolchainManagement.swift
@@ -54,6 +54,14 @@ extension FileSystem {
   private var cartonSDKPath: AbsolutePath {
     homeDirectory.appending(components: ".carton", "sdk")
   }
+  
+  public var swiftVersionPath: AbsolutePath {
+    guard let cwd = currentWorkingDirectory else {
+      fatalError()
+    }
+    
+    return cwd.appending(component: ".swift-version")
+  }
 
   func inferSwiftVersion(
     from versionSpec: String? = nil,
@@ -70,16 +78,9 @@ extension FileSystem {
       }
     }
 
-    guard let cwd = currentWorkingDirectory else { return defaultToolchainVersion }
-
-    let versionFile = cwd.appending(component: ".swift-version")
-
-    guard isFile(versionFile), let version = try readFileContents(versionFile)
-      .validDescription?
-      // get the first line of the file
-      .components(separatedBy: CharacterSet.newlines).first,
-      version.contains("wasm")
-    else { return defaultToolchainVersion }
+    guard let version = try fetchLocalSwiftVersion(), version.contains("wasm") else {
+      return defaultToolchainVersion
+    }
 
     return version
   }
@@ -290,5 +291,20 @@ extension FileSystem {
     }
 
     return result.sorted()
+  }
+  
+  public func fetchLocalSwiftVersion() throws -> String? {
+    guard isFile(swiftVersionPath), let version = try readFileContents(swiftVersionPath)
+      .validDescription?
+      // get the first line of the file
+      .components(separatedBy: CharacterSet.newlines).first
+    
+    else { return nil }
+    
+    return version
+  }
+  
+  public func setLocalSwiftVersion(_ version: String) throws {
+    try writeFileContents(swiftVersionPath, bytes: ByteString([UInt8](version.utf8)))
   }
 }
--- a/Sources/SwiftToolchain/ToolchainManagement.swift
+++ b/Sources/SwiftToolchain/ToolchainManagement.swift
@@ -298,7 +298,6 @@ extension FileSystem {
       .validDescription?
       // get the first line of the file
       .components(separatedBy: CharacterSet.newlines).first
-    
     else { return nil }
     
     return version
--- a/Sources/carton/Commands/Local.swift
+++ b/Sources/carton/Commands/Local.swift
@@ -1,0 +1,48 @@
+// Copyright 2020 Carton contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


Sources/SwiftToolchain/ToolchainManagement.swift
```diff
@@ -291,4 +284,27 @@ extension FileSystem {
...
+  public func fetchLocalSwiftVersion() throws -> String? {
+    guard let cwd = currentWorkingDirectory else { return nil }
+
+    let versionFile = cwd.appending(component: ".swift-version")
+
+    guard isFile(versionFile), let version = try readFileContents(versionFile)
+      .validDescription?
+      // get the first line of the file
+      .components(separatedBy: CharacterSet.newlines).first,
+      version.contains("wasm")
```
MaxDesiatov, line 297: Please make `fetchLocalSwiftVersion()` return the version as is, without any checks for the `wasm` substring:

```suggestion
      .components(separatedBy: CharacterSet.newlines).first
```

Sources/SwiftToolchain/ToolchainManagement.swift
```diff
@@ -70,16 +70,9 @@ extension FileSystem {
...
-
-    let versionFile = cwd.appending(component: ".swift-version")
-
-    guard isFile(versionFile), let version = try readFileContents(versionFile)
-      .validDescription?
-      // get the first line of the file
-      .components(separatedBy: CharacterSet.newlines).first,
-      version.contains("wasm")
-    else { return defaultToolchainVersion }
+    guard let version = try fetchLocalSwiftVersion() else {
```
MaxDesiatov, line 73: If the function is updated as in the suggestion below, the substring check would make sense here:

```suggestion
    guard 
      let version = try fetchLocalSwiftVersion(), 
      version.contains("wasm") 
    else {
```

Sources/SwiftToolchain/ToolchainManagement.swift
```diff
@@ -291,4 +284,27 @@ extension FileSystem {
 
     return result.sorted()
   }
+  
+  public func fetchLocalSwiftVersion() throws -> String? {
+    guard let cwd = currentWorkingDirectory else { return nil }
+
+    let versionFile = cwd.appending(component: ".swift-version")
```
MaxDesiatov, line 291: Please generalize this as `var versionFilePath: AbsolutePath` computed property, similarly to how it was done for `swiftenvVersionsPath` and `cartonSDKPath` above. It would be fair to `fatalError` if `currentWorkingDirectory` returns `nil`, this will allow making the return type of `versionFilePath` non-optional.

Sources/SwiftToolchain/ToolchainManagement.swift
```diff
@@ -291,4 +284,27 @@ extension FileSystem {
...
+    
+    return version
+  }
+  
+  public func setLocalSwiftVersion(_ version: String) throws {
+    guard let cwd = currentWorkingDirectory else { return }
+    
+    let versionFile = cwd.appending(component: ".swift-version")
+    
+    try writeFileContents(versionFile, bytes: ByteString([UInt8](version.utf8)))
```
MaxDesiatov, line 308: After `versionFilePath` property is created, it can be shorted to this:

```suggestion
    try writeFileContents(versionFilePath, bytes: ByteString([UInt8](version.utf8)))
```

Sources/carton/Commands/Local.swift
```diff
@@ -0,0 +1,45 @@
...
+  
+  @Argument() var version: String?
+  
+  func run() throws {
+    guard let terminal = TerminalController(stream: stdoutStream) else {
+      fatalError("failed to create an instance of `TerminalController`")
+    }
+    
+    guard let localVersion = try localFileSystem.fetchLocalSwiftVersion() else {
+      terminal.write("You are not in a Swift project!")
```
MaxDesiatov, line 29: **suggestion**: I think this message is somewhat misleading, you certainly can navigate to a Swift project without a `.swift-version` file, would be more useful to log what actually is missing:

```suggestion
      terminal.logLookup("Version file is not present: ", localFileSystem.versionFilePath)
```

Sources/carton/Commands/Local.swift
```diff
@@ -0,0 +1,45 @@
...
+    guard let version = version else {
+      terminal.write("\(localVersion)", inColor: .green)
+      return
+    }
+    
+    let versions = try localFileSystem.fetchAllSwiftVersions()
+    if (versions.contains(version)) {
+      _ = try localFileSystem.setLocalSwiftVersion(version)
+    } else {
+      terminal.write("The version hasn't been installed!", inColor: .red)
```
MaxDesiatov, line 42: **suggestion:** let's make error message as explicit as possible.

```suggestion
      terminal.write("The version "\(version)" hasn't been installed!", inColor: .red)
```

Sources/carton/Commands/Local.swift
```diff
@@ -0,0 +1,45 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArgumentParser
+import TSCBasic
+
+struct Local: ParsableCommand {
+  static var configuration = CommandConfiguration(abstract: "")
```
MaxDesiatov, line 19: Please add an abstract string for this subcommand:

```suggestion
  static var configuration = CommandConfiguration(
    abstract: """
    Prints SDK version used for the current project or saves it \
    in the `.swift-version` file if a version is passed as an argument.
    """
  )
```

RayZhao1998:
Thanks for your comments. I'm a little busy yesterday and I've fixed all the suggestions today. Please review it again.

Sources/SwiftToolchain/ToolchainManagement.swift
```diff
@@ -291,4 +292,19 @@ extension FileSystem {
 
     return result.sorted()
   }
+  
+  public func fetchLocalSwiftVersion() throws -> String? {
+    guard isFile(swiftVersionPath), let version = try readFileContents(swiftVersionPath)
+      .validDescription?
+      // get the first line of the file
+      .components(separatedBy: CharacterSet.newlines).first
+
```
MaxDesiatov, line 301: **nitpick:** no need for the empty line here, which would be consistent with the rest of `guard` statements in the codebase
```suggestion
```

PR #46, opened 2020-07-14
Add sources to watcher per target

### LINKED ISSUE #1: Watcher should detect custom paths in Package.swift
Currently, the watcher defaults to the `Sources` directory for the `dev` command and `Sources` and `Tests` directories for the `test` command. If the source code is in a different location overriden in target definitions in `Package.swift`, the watcher doesn't pick those up.

JaapWijnen:
Sources are added per target and use a custom path if provided by the package manifest. This should fix #1 !

### DIFF
--- a/Sources/SwiftToolchain/Toolchain.swift
+++ b/Sources/SwiftToolchain/Toolchain.swift
@@ -107,7 +107,7 @@ public final class Toolchain {
     }
   }
     
-  public func inferSourcePaths() throws -> [String] {
+  public func inferSourcesPaths() throws -> [String] {
     let package = try Package(with: swiftPath, terminal)
     
     let targetPaths = package.targets.compactMap { target -> String? in
--- a/Sources/SwiftToolchain/Toolchain.swift
+++ b/Sources/SwiftToolchain/Toolchain.swift
@@ -110,13 +110,13 @@ public final class Toolchain {
   public func inferSourcePaths() throws -> [String] {
     let package = try Package(with: swiftPath, terminal)
     
-    let targetPaths = package.targets.map { target -> String in
+    let targetPaths = package.targets.compactMap { target -> String? in
         guard let path = target.path else {
             switch target.type {
             case .regular:
                 return "Sources/\(target.name)"
             case .test:
-                return "Tests/\(target.name)"
+                return nil
             }
         }
         return path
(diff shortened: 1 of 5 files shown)


Sources/SwiftToolchain/Toolchain.swift
```diff
@@ -106,6 +106,24 @@ public final class Toolchain {
...
+  public func inferSourcePaths() throws -> [String] {
+    let package = try Package(with: swiftPath, terminal)
+    
+    let targetPaths = package.targets.map { target -> String in
+        guard let path = target.path else {
+            switch target.type {
+            case .regular:
+                return "Sources/\(target.name)"
+            case .test:
+                return "Tests/\(target.name)"
```
MaxDesiatov, line 119: **suggestion:** not sure we should watch test paths, at least not in this function, which is called from `carton dev`. `carton test` would probably need to call a separate function to watch both sources and tests, but just watching sources here should be enough.

```suggestion
    let targetPaths = package.targets.compactMap { target -> String? in
        guard let path = target.path else {
            switch target.type {
            case .regular:
                return "Sources/\(target.name)"
            case .test:
                return nil
```

Sources/SwiftToolchain/Toolchain.swift
```diff
@@ -106,6 +106,24 @@ public final class Toolchain {
       return nil
     }
   }
+    
+  public func inferSourcePaths() throws -> [String] {
```
MaxDesiatov, line 110: **nitpick:** let's name it consistently with the thing being watched.

```suggestion
  public func inferSourcesPaths() throws -> [String] {
```

Sources/carton/Commands/Dev.swift
```diff
@@ -49,23 +49,30 @@ struct Dev: ParsableCommand {
...
     )
-
-    guard let sources = localFileSystem.currentWorkingDirectory?.appending(component: "Sources")
-    else { fatalError("failed to infer the sources directory") }
-
-    terminal.write("\nWatching this directory for changes: ", inColor: .green)
-    terminal.logLookup("", sources)
-    terminal.write("\n")
+    
+    let sources = try toolchain.inferSourcePaths().map { source -> [AbsolutePath] in
```
MaxDesiatov, line 61: **nitpick**


```suggestion
    let sources = try toolchain.inferSourcesPaths().map { source -> [AbsolutePath] in
```

JaapWijnen:
Thanks for the comments! committed the changes :)

MaxDesiatov:
Brilliant 👏

PR #61, opened 2020-07-20
Propagate test build/run failures in the exit code

### LINKED ISSUE #56: Avoid running the tests if can't build them
`carton test` currently runs old test bundle binaries if present, if building a new test bundle failed.

MaxDesiatov:
Resolves #56.

Also, a brief doc comment is added to the `Equality` assertions helper type.

### DIFF
--- a/Sources/CartonHelpers/ProcessRunner.swift
+++ b/Sources/CartonHelpers/ProcessRunner.swift
@@ -21,6 +21,17 @@ import OpenCombine
 #endif
 import TSCBasic
 
+extension Subscribers.Completion {
+  var result: Result<(), Failure> {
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
Thank you for the review and welcome to the `carton` maintainers team! I'll wait for @carson-katri's review a bit more, as I'd like the next release of `carton` to coincide with the imminent 0.2.0 release of `Tokamak`. That would allow `carton init --template tokamak` to specify 0.2.0 as a package dependency instead of the `main` branch as it does now. And `Tokamak` still would benefit from a few currently open PRs to be merged before the release, so after all there's no need to rush this one.

PR #67, opened 2020-07-23
Fix watcher crashing in package subdirectories

MaxDesiatov:
Currently, when `carton dev` is launched in any subdirectory of the package not containing `Package.swift` it crashes. It's caused by `traverseRecursively` precondition failing when a non-directory (or non-existent) path is passed to it. I think `traverseRecursively` can handle non-existing paths well by returning an empty array, and the path itself when it's a file and not a directory. Thus the precondition check is not needed at all.

Non-existent paths were created due to a wrong assumption that the root package directory is always the current directory. This is now fixed by adding a new `inferManifestDirectory` function, which correctly calculates absolute paths of source directories before passing them to the watcher.

Additionally the watcher code in the `TSCBasic` code has its own precondition for non-empty path arrays passed to it. To avoid triggering it and staying extra-safe, we check for an empty watcher paths array and avoid watching anything at all.

### DIFF
--- a/Sources/SwiftToolchain/Toolchain.swift
+++ b/Sources/SwiftToolchain/Toolchain.swift
@@ -126,9 +126,6 @@ public final class Toolchain {
       throw ToolchainError.missingPackage
     }
 
-    // FIXME: this won't work on Windows
-    let root = AbsolutePath("/")
-
     repeat {
       guard !fileSystem.isFile(cwd.appending(component: "Package.swift")) else {
         return cwd
@@ -136,7 +133,7 @@ public final class Toolchain {
 
       // `parentDirectory` just returns `self` if it's `root`
       cwd = cwd.parentDirectory
-    } while cwd != root
+    } while !cwd.isRoot
 
     throw ToolchainError.missingPackage
   }
--- a/Sources/SwiftToolchain/Toolchain.swift
+++ b/Sources/SwiftToolchain/Toolchain.swift
@@ -44,7 +44,10 @@ enum ToolchainError: Error, CustomStringConvertible {
     case .failedToBuildTestBundle:
       return "Failed to build the test bundle"
     case .missingPackage:
-      return "No package found"
+      return """
+      The `Package.swift` manifest file could not be found. Please navigate to a directory that \
+      contains `Package.swift` and restart.
+      """
     }
   }
 }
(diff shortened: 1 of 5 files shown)


Sources/SwiftToolchain/Toolchain.swift
```diff
@@ -118,8 +121,30 @@ public final class Toolchain {
     }
   }
 
+  private func inferManifestDirectory() throws -> AbsolutePath {
+    guard package != nil, var cwd = fileSystem.currentWorkingDirectory else {
+      throw ToolchainError.missingPackage
+    }
+
+    // FIXME: this won't work on Windows
```
ie-ahm-robox, line 129: - :no_entry_sign: FIXMEs should be resolved (this won't work on Windows). (`todo`)

Sources/carton/Helpers/FileSystem.swift
```diff
@@ -16,11 +16,18 @@ import Foundation
...
+    }
+
+    var result = [traversalRoot]
+
+    guard isDirectory(traversalRoot) else {
+      return result
+    }
 
-    var pathsToTraverse = [root]
+    var pathsToTraverse = [traversalRoot]
```
RayZhao1998, line 30:
```suggestion
    var pathsToTraverse = result
```

Sources/SwiftToolchain/Toolchain.swift
```diff
@@ -118,8 +121,30 @@ public final class Toolchain {
...
+    let root = AbsolutePath("/")
+
+    repeat {
+      guard !fileSystem.isFile(cwd.appending(component: "Package.swift")) else {
+        return cwd
+      }
+
+      // `parentDirectory` just returns `self` if it's `root`
+      cwd = cwd.parentDirectory
+    } while cwd != root
```
RayZhao1998, line 139: It seems that `AbsolutePath` has a property `isRoot` to check whether it's a root directory.
```suggestion
    } while !cwd.isRoot
```

Sources/carton/Helpers/FileSystem.swift
MaxDesiatov: Good catch, thanks!

Sources/SwiftToolchain/Toolchain.swift
ie-ahm-robox, line 129: - :no_entry_sign: FIXMEs should be resolved (this won't work on Windows). (`todo`)

ie-ahm-robox, line 129: - :no_entry_sign: FIXMEs should be resolved (this won't work on Windows). (`todo`)

Sources/SwiftToolchain/Toolchain.swift
MaxDesiatov: Great point, I initially wasn't sure if this property is public at all.

PR #69, opened 2020-07-25
Add verbose/concise logging levels

carson-katri:
This adds a `--verbose/-v` flag to `carton dev`.

When enabled, logging acts the same as before this PR.

Otherwise, the screen is cleared before starting, and after every rebuild. This keeps the most recent diagnostics visible, while keeping the terminal relatively clean. It also switches the Vapor `Environment` to `"production"` to only show logs >= notices.

`-Xswiftc -color-diagnostics` is also passed to the Swift compiler for both levels.

Let me know if any of this is behavior isn't desired.

### DIFF
--- a/Sources/SwiftToolchain/Toolchain.swift
+++ b/Sources/SwiftToolchain/Toolchain.swift
@@ -209,6 +209,7 @@ public final class Toolchain {
     let builderArguments = try [
       swiftPath.pathString, "build", "-c", isRelease ? "release" : "debug", "--build-tests",
       "--destination", inferDestinationPath().pathString,
+      "-Xswiftc", "-color-diagnostics",
     ]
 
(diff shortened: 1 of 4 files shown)


Sources/SwiftToolchain/Toolchain.swift
```diff
@@ -176,6 +176,7 @@ public final class Toolchain {
     let builderArguments = try [
       swiftPath.pathString, "build", "-c", isRelease ? "release" : "debug", "--product", product,
       "--enable-test-discovery", "--destination", destination ?? inferDestinationPath().pathString,
+      "-Xswiftc", "-color-diagnostics",
```
MaxDesiatov, line 179: Please also add these arguments to the invocation in `buildTestBundle` below.

PR #86, opened 2020-08-17
Propagate package manifest parsing errors

### LINKED ISSUE #78: Print the error output of `swift package dump-package`
Currently when `swift package dump-package` fails, its error output is not presented to the user, while the printed error message is confusing. In 0.4.1 it's 
```
Error: No package found
```
while in `main` it's 
```
Error: The `Package.swift` manifest file could not be found. Please
navigate to a directory that contains `Package.swift` and restart
```

Neither of these are valid error messages when `Package.swift` is present, but it has a typo or doesn't compile for some other reason.

MaxDesiatov:
Previously, initial manifest parsing error was ignored with `try?` and `ToolchainError.missingPackageManifest` was thrown instead. Now the `package` property is no longer optional, but is of `Result` type, where the package manifest parsing error is stored and re-thrown later.

Resolves #78.

### DIFF
--- a/Sources/SwiftToolchain/Toolchain.swift
+++ b/Sources/SwiftToolchain/Toolchain.swift
@@ -188,8 +188,7 @@ public final class Toolchain {
     let package = try self.package.get()
 
     if let jsKit = package.dependencies?.first(where: { $0.name == "JavaScriptKit" }),
-      jsKit.requirement.revision != ["c90e82f"]
-    {
+      jsKit.requirement.revision != ["c90e82f"] {
       let version = jsKit.requirement.revision.flatMap { " (\($0[0]))" } ?? ""
 
       terminal.write(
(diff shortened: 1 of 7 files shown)


Sources/SwiftToolchain/Toolchain.swift
```diff
@@ -177,9 +185,11 @@ public final class Toolchain {
...
     else { throw ToolchainError.noExecutableProduct }
 
-    if let package = package,
-      let jsKit = package.dependencies?.first(where: { $0.name == "JavaScriptKit" }),
-      jsKit.requirement.revision != ["c90e82f"] {
+    let package = try self.package.get()
+
+    if let jsKit = package.dependencies?.first(where: { $0.name == "JavaScriptKit" }),
+      jsKit.requirement.revision != ["c90e82f"]
+    {
```
ie-ahm-robox, line 192: - :no_entry_sign: Opening braces should be preceded by a single space and on the same line as the declaration. (`opening_brace`)

MaxDesiatov:
I've removed unused `ToolchainError` case, which I initially added, but didn't end up using.

RayZhao1998 (approved):

PR #88, opened 2020-08-17
Fix broken `carton init` templates

### LINKED ISSUE #77: JavaScriptKit dependency missing in the `basic` template
While `carton` ships JavaScriptKit runtime, the runtime bits don't work without the Swift bits of JavaScriptKit. The default `basic` template has no `JavaScriptKit` dependency specified in the generated `Package.swift`, which leads to runtime errors when a project is built from this template as is.

MaxDesiatov:
Resolves #77.

Also updates the Tokamak template to use the `App`/`Scene` lifecycle and to depend on the 0.3 version.

JavaScriptKit version check is also updated.


MaxDesiatov:
Hey @carson-katri @RayZhao1998, would you mind having a look at this please? I'm going to be away probably most of next week, but would be amazing if this could be released and announced together with the new release of Tokamak to a wide audience tomorrow before I'm off. The change is relatively small and I don't think too disruptive, so should be an easy review. Thanks!

RayZhao1998 (approved):

PR #97, opened 2020-08-28
Implement `carton bundle` command

### LINKED ISSUE #16: Implement `carton bundle` command
This command should invoke `swift build` (reusing the toolchain inference code that's already used by `carton dev` and `carton sdk`) in the directory for the only executable product, or a product specified with the `--product` option, similar to how `carton dev` infers a product to build. It should build the app in the release mode with `swift build -c release`, possibly passing `-Osize` flag to `swiftc` if it the release configuration doesn't do this by default already. After that, it should look for the presence of `wasm-strip` and `wasm-opt` on the system and apply size optimizations with those if present, otherwise suggesting to install corresponding [WABT](https://github.com/WebAssembly/wabt) and [binaryen](https://github.com/WebAssembly/binaryen) packages (via Homebrew on macOS).

We also need a new `bundle.js` JavaScript entrypoint, similar to what we have with `dev.js`, but without the WebSocket bits, as the release bundle doesn't need hot reloads. This entrypoint should also by compiled by `webpack` in prod mode for size optimizations. The only question I don't have a clear answer to is, what the final bundle directory should be to which `carton bundle` writes the output? A few suggestions:

* `.build/bundle/wasm`, means that the generated fails get ignored by Git most probably (since `.build` is in `.gitignore` most of the time)
* `Bundle`, capitalized for consistency with `Sources` and `Tests`
* `CartonBundle`, in case `Bundle` is used by other tools, but this one seems a bit verbose to me.

I have a slight preference towards the first option as these files are not supposed to be added to the repository in most of the cases, but the problem is that `.build` is not visible in Finder and other file managers by default, so I could see some beginner users not finding the final product on the file system. With directories like `Bundle` we'd need users to explicitly opt out of committing the final products to their repository (again, most of the time, unless they don't have `.build` in their `.gitignore` yet, which would make this point applicable to the `.build` option too).

MaxDesiatov:
New `bundle.js` entrypoint is added, which only differs from `dev.js` in the lack of the WebSocket hot reloading bit.

`wabt` and `binaryen` Homebrew dependencies are added as required for `wasm-strip` and `wasm-opt` respectively that reduce the resulting bundle binary size.

All resulting bundle files except `index.html` are named by their content hashes to enable [cache busting](https://www.keycdn.com/support/what-is-cache-busting).

Resolves #16.


Sources/carton/Helpers/FileSystem.swift
```diff
@@ -38,4 +38,11 @@ extension FileSystem {
 
     return result
   }
+
+  func humanReadableFileSize(_ path: AbsolutePath) throws -> String {
+    precondition(isFile(path))
+
+    // FIXME: should use `UnitInformationStorage`, but it's unavailable in open-source Foundation
```
ie-ahm-robox, line 45: - :no_entry_sign: FIXMEs should be resolved (should use `UnitInformationSto...). (`todo`)

ie-ahm-robox:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  **Sources/carton/Model/Dependency.swift#L42** - Tuples should have at most 2 members. (`large_tuple`)
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against 489342dfb6dd83503ad22e3453762d83004ff865
</p>

Sources/carton/Helpers/FileSystem.swift
ie-ahm-robox, line 45: - :no_entry_sign: FIXMEs should be resolved (should use `UnitInformationSto...). (`todo`)

ie-ahm-robox, line 45: - :no_entry_sign: FIXMEs should be resolved (should use `UnitInformationSto...). (`todo`)

PR #104, opened 2020-09-04
Implement support for static resources in `carton dev`

### LINKED ISSUE #38: Add static file support
This is a prerequisite for https://github.com/swiftwasm/Tokamak/pull/155. Ideally there would be support for using `Bundle` to get the URL for the static resources at runtime.

---

@MaxDesiatov:
> I think that both `carton` and Tokamak should support [SwiftPM resources](https://github.com/apple/swift-evolution/blob/master/proposals/0271-package-manager-resources.md), the only obvious problem being that the `Bundle` type is declared in Foundation...

@j-f1:
> Could `carton` or something else implement the `Bundle` class itself?

@MaxDesiatov:
> It probably could, but I'm not sure whether SwiftPM assumes a specific `Bundle` implementation in its code and whether we could swap that with our own. Needs some testing and research... 🧪

MaxDesiatov:
This change requires `wasm-5.3-SNAPSHOT-2020-09-08-a` or later, therefore the default toolchain version has been bumped.

Implements the `carton dev` part of #38, `carton bundle` support will be implemented in a future PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ is rarely needed, as `carton dev` install the recommended version of SwiftWasm a
 `carton sdk versions` lists all installed versions, and `carton sdk local` prints the version
 specified for the current project in the `.swift-version` file. You can however install SwiftWasm
 separately if needed, either by passing an archive URL to `carton sdk install` directly, or just
-specifying the snapshot version, like `carton sdk install wasm-DEVELOPMENT-SNAPSHOT-2020-06-07-a`.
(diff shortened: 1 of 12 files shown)


MaxDesiatov:
@swiftwasm/carton-team would anyone be able to have a look at this? Please lmk if you'd like me to go ahead and merge PRs that don't get a review more than a week in general, or if we should have some other approach for this.

j-f1 (approved):
LGTM!

RayZhao1998 (approved):

PR #113, opened 2020-09-26
Use raw stdout if TerminalController is unavailable

### LINKED ISSUE #112: `carton` crashes when it fails to instantiate `TerminalController`
This is reproducible when you run any of the `carton` commands on CI, such as GitHub Actions.

MaxDesiatov:
TTY terminal output is unavailable on CI, or when users redirect output to a file or a pipe. In those cases `TerminalController` initializer returns `nil`. SwiftPM has a wrapper `InteractiveWriter` class for it, which writes output to stdout directly in that case. I've copied that, and replaced all uses of `TerminalController` with it. A VSCode task is added to test that.

In my testing it looks like download progress reporting is quite spammy, so it now has `removeDuplicates` Combine operator added to make it output only when values differ significantly from each other.

Additionally, SDK installs can hit GitHub API rate limit on CI nodes, so `carton` now reads `GITHUB_TOKEN` environment variable, which gives much higher limits to authenticated API users.

We can now also run various `carton` commands on CI for basic end-to-end testing.

Resolves #112.


ie-ahm-robox:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  **Sources/SwiftToolchain/ToolchainManagement.swift#L268** - Function body should span 50 lines or less excluding comments and whitespace: currently spans 54 lines (`function_body_length`)
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against 7dfbef3256b6401c4fd94bb8ecb5102690b48e5b
</p>

carson-katri (approved):

PR #116, opened 2020-10-01
Allow changing dev server's port

yonihemi:
Running `carton-release` also changed bundle.js's hash, so I just didn't commit the static resources.

### DIFF
--- a/entrypoint/dev.js
+++ b/entrypoint/dev.js
@@ -34,7 +34,7 @@ const wasi = new WASI({
 });
 
 const port = document.currentScript.getAttribute("data-port") || 8080;
-const socket = new ReconnectingWebSocket(`ws://127.0.0.1:${port}/watcher`);
+const socket = new ReconnectingWebSocket(`ws://${location.host}/watcher`);
 
 socket.addEventListener("message", (message) => {
   if (message.data === "reload") {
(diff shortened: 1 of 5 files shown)


entrypoint/dev.js
```diff
@@ -33,7 +33,8 @@ const wasi = new WASI({
   },
 });
 
-const socket = new ReconnectingWebSocket("ws://127.0.0.1:8080/watcher");
+const port = document.currentScript.getAttribute("data-port") || 8080;
+const socket = new ReconnectingWebSocket(`ws://127.0.0.1:${port}/watcher`);
```
j-f1, line 37: This should make passing the port unnecessary.
```suggestion
const socket = new ReconnectingWebSocket(`ws://${location.host}/watcher`);
```

MaxDesiatov: I concur, let's use `location.host`, which means that passing the port to `indexPage` and `data-port` attribute won't be needed.

MaxDesiatov (changes requested):
This is a great feature and I would love to merge this, but I think it would be simpler to use `location.host` as Jed suggested in his comment.

MaxDesiatov (approved):
Wonderful, thank you!

j-f1 (approved):

RayZhao1998 (approved):

PR #131, opened 2020-10-16
Add separate Builder class, use WasmTransformer

### LINKED ISSUE #127: Method with i64 return type fails on Safari 13+14
Using JavaScriptKit as only dependency, trying to instantiate a `Date()` fails on Safari (13.1.1, 14) with:
**TypeError: i64 not allowed as return type or argument to an imported function**
<details>
<summary>Stacktrace</summary>
<p>

```
TypeError: i64 not allowed as return type or argument to an imported function

wasm-stub
<?>.wasm-function[gettimeofday]
<?>.wasm-function[CFAbsoluteTimeGetCurrent]
<?>.wasm-function[$s10Foundation4DateVACycfC]
<?>.wasm-function[main]
<?>.wasm-function[__original_main]
<?>.wasm-function[_start]
wasm-stub
_start
(anonymous function) — e8900d1b3895f276.js:1:47960
(anonymous function) — e8900d1b3895f276.js:1:184568
asyncFunctionResume
[native code]
```
</p>
</details>

Same code runs fine on Chrome and Firefox. Tried with latest blessed carton versions and with latest swiftwasm+JavaScriptKit, same results.

Assuming it's related to swiftwasm/JavaScriptKit#56. Could/should we incorporate this polyfill in JavaScriptKit?

MaxDesiatov:
Resolves #127.

### DIFF
--- a/Sources/SwiftToolchain/Builder.swift
+++ b/Sources/SwiftToolchain/Builder.swift
@@ -23,10 +23,16 @@ import TSCBasic
 import WasmTransformer
 
 public final class Builder {
+  public enum Environment {
+    case wasmer
+    case browser
+  }
+
   public let mainWasmPath: AbsolutePath
 
   private var currentProcess: ProcessRunner?
   private let arguments: [String]
+  private let environment: Environment
   private let terminal: InteractiveWriter
   private let fileSystem: FileSystem
   private var subscription: AnyCancellable?
@@ -34,11 +40,13 @@ public final class Builder {
   public init(
     arguments: [String],
     mainWasmPath: AbsolutePath,
+    environment: Environment = .browser,
     _ fileSystem: FileSystem,
     _ terminal: InteractiveWriter
   ) {
     self.arguments = arguments
     self.mainWasmPath = mainWasmPath
+    self.environment = environment
     self.terminal = terminal
     self.fileSystem = fileSystem
   }
@@ -57,6 +65,8 @@ public final class Builder {
           "`swift build` completed in ",
           String(format: "%.2f seconds", abs(buildStarted.timeIntervalSinceNow))
         )
+
+        guard self.environment == .browser else { return }
 
         // FIXME: errors from these `try` expressions should be recoverable, not sure how to
         // do that in `handleEvents`, and `flatMap` doesnt' fit here as we need to track
--- a/Sources/SwiftToolchain/Builder.swift
+++ b/Sources/SwiftToolchain/Builder.swift
@@ -1,0 +1,90 @@
+// Copyright 2020 Carton contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import CartonHelpers
+#if canImport(Combine)
+import Combine
+#else
+import OpenCombine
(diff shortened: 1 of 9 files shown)


kateinoigakukun:
I want to note that if a binary is lowered into i32, it can't be used on some wasi compatible runtime like wasmtime. So the lowering should run only when targeting browser.

Sources/SwiftToolchain/Builder.swift
```diff
@@ -0,0 +1,100 @@
...
+#else
+import OpenCombine
+#endif
+import Foundation
+import TSCBasic
+import WasmTransformer
+
+public final class Builder {
+  public enum Environment {
+    case wasmer
```
kateinoigakukun, line 27: We should consider two environments, browser and others. I think Wasmer is very specific environment. We should consider in more general.

MaxDesiatov: I see these becoming more specific in the future, even mentioning specific browsers with 

```swift
case firefox(version: Int)
case safari(version: Int)
case chrome(version: Int)
```

This would allow `carton` to determine which polyfills to embed and what transformations to apply depending on what this specific browser supports. Similarly `wasmer` is specific enough I think. I'd add `nodeJS` in the near future, which for example would require shipping WASI polyfills for it.

MaxDesiatov: Can add `case unknownWasmHost` for now, which would denote hosts that don't have WASI embedded, but it would stay unused right now anyway as we do require WASI.

kateinoigakukun: This i64 issue is not limited to wasi. I think it's better to allow users to give an option to indicate that the target environment supports i64 imports or not.

e.g.

```
$ carton build --lower-i64-imports
```

MaxDesiatov: I'm a bit worried about the amount of command-line options we've already accumulated. If we decide to expose every transformation as an option, this would make the list even bigger, and even harder to maintain. The main philosophy of `carton` is to be easy as possible to use and infer configuration settings automatically. This would allow users to specify something like `--environment Safari` in the future, and `carton` should infer all build settings automatically from it. If users need any customizations, they should either open an issue and let us know what and why is needed, or use webpack.js or something else if we can't satisfy their needs.

We may add support for config files, but I would want those to be completely optional, so that users just install `carton` and get the best experience out of the box without any configs.

At least that's my current thinking, I'm happy to hear any feedback 🙂

MaxDesiatov: Another thing to note is that `bundle`, `dev` and `test` commands are applicable to other platforms too, such as Android, Linux Windows. I hope we can support cross-compiling (or even directly compiling for the host platform in case of Windows) to these platforms in the future. So if we add more platform-specific command-line options, we're setting a precedent here for more command-line options for those platforms too. As I mentioned above, I don't think that's tenable long-term. I hope people would just specify `carton test --environment Safari` or `carton test --environment AndroidSimulator` and it would sort things out for them by picking the best options.

MaxDesiatov: @kateinoigakukun Sorry, regardless of what I said above, I realize now that `case wasmer` is very specific, and `case browser` is very general. You're right, `case other` makes total sense here. I suggest we discuss the addition of CLI flags separately, and this one is ready for review 🙂

kateinoigakukun: I think if we want carton to be simple, it's better to focus on targeting only browser environment and interoperation around JavaScript.

kateinoigakukun (approved):

Sources/SwiftToolchain/Builder.swift
MaxDesiatov: There's also a toolchain downloader, which I think is a big feature too, especially as it simplifies the installation process for newcomers. I wish something in the Swift toolchain could do this, but it doesn't look like upstream developers are interested in this.

carson-katri (approved):

PR #134, opened 2020-10-19
Fix support for Ubuntu 20.04, use GHA for SwiftLint

### LINKED ISSUE #114: Support downloading Ubuntu 20.04 SDK
We currently don't check for distribution version on Linux, which means we could run on unsupported distributions, or download wrong toolchain that won't work. `carton` should check which Linux distribution it runs on, and then download a compatible toolchain archive for it, or notify the user about incompatibility.

Accordingly, it should check CPU architecture, as it would be great to support SwiftWasm on ARM64 Linux distributions, e.g. Raspbian after https://github.com/swiftwasm/swift/issues/1854 is resolved.

MaxDesiatov:
Resolves #114.

As support for SwiftLint in Danger seems a bit buggy (causing duplicate warning comments sometimes), this PR calls SwiftLint through a separate action that displays warnings only in the PR diff and hopefully in better way.


ie-ahm-robox:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  **Sources/SwiftToolchain/ToolchainManagement.swift#L404** - File should contain 400 lines or less: currently contains 404 (`file_length`)
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against 570ce9644a5e0639d8822f39caa9c958684a25c6
</p>

carson-katri (dismissed):

yonihemi (approved):

PR #140, opened 2020-10-23
Update JavaScriptKit version in TestApp

kateinoigakukun:


MaxDesiatov:
Maybe it's worth updating the test app code too to use the new API without force unwrapping?

yonihemi (approved):

MaxDesiatov (approved):
Great, thanks 👍

PR #141, opened 2020-10-23
Stop using destination.json

kateinoigakukun:
After https://github.com/swiftwasm/swift/pull/2081, we don't need to include non static `lib/swift` directory to include Foundation and other system modules. We should tell the driver to include module and link libraries as static mode.


kateinoigakukun:
FYI: I'm working to remove use of `destination.json` to avoid complexity.

- https://github.com/swiftwasm/swift/pull/2081 : This enables to remove `-I toolchain/path/lib/swift/wasi/wasm32`
- https://github.com/swiftwasm/swift/pull/2089 : This enables to remove `sdk` section
- https://github.com/swiftwasm/swift-corelibs-foundation/pull/156 : This enables to remove all linker flags

kateinoigakukun:
macOS test fail until https://github.com/swiftwasm/swift-package-manager/pull/366

kateinoigakukun:
Now, ready for review

TestApp/Sources/TestApp/main.swift
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import CoreFoundation
```
yonihemi, line 15: Should we consider keeping this as an extraSwiftcFlag? Won't be surprised if this deters a lot familiar-to-Swift users who are new to SwiftWasm.

kateinoigakukun: Users should import `CoreFoundation` explicitly because Foundation imports it with `@_implementationOnly` which means users can choose to link it or not. Currently, there is [an import](https://github.com/apple/swift-corelibs-foundation/blob/6787b8a9912a392234062ff3e1be978acc727d29/Sources/Foundation/RunLoop.swift#L11) without the attribute, so no other platform toolchain doesn't require to import CoreFoundation now, but that only one import statement will be removed in the future.

If a user doesn't use any Foundation API that depends on CoreFoundation, he doesn't need to link CoreFoundation. It results in reducing output binary size. If we always link CoreFoundation using extra flags, the output binary size will be more fat. (libCoreFoundation.a is 3.6 MB)

However,  `@_implementationOnly` attribute is not applied in 5.3 branch, so explicit import is required only when using DEVELOPMENT toolchain.

MaxDesiatov (dismissed):
Super happy to see that explicit destinations are gone now. Thank you very much for cleaning it up!

Sources/carton/Commands/Dev.swift
```diff
@@ -32,9 +32,6 @@ struct Dev: ParsableCommand {
   @Option(help: "Specify name of an executable product in development.")
   var product: String?
 
-  @Option(help: "Specify name of a json destination file to be passed to `swift build`.")
```
MaxDesiatov, line 35: Just a suggestion here, let's keep the flag and notify users that this option no longer has any effect and will be removed in a future version. This will give some transition period to whoever was still using it. WDYT?

Sources/carton/Commands/Dev.swift
```diff
@@ -32,7 +32,9 @@ struct Dev: ParsableCommand {
...
+  @Option(
+    help: "Specify name of a json destination file to be passed to `swift build`. (deprecated)"
```
MaxDesiatov, line 36:
```suggestion
    help: "This option no longer has any effect and will be removed in a future version of `carton`"
```

Sources/carton/Commands/Dev.swift
```diff
@@ -66,9 +68,12 @@ struct Dev: ParsableCommand {
       terminal.saveCursor()
     }
 
+    if destination != nil {
+      terminal.write("--destination is no longer supported.\n", inColor: .yellow)
```
MaxDesiatov, line 72: **suggestion:** I think it should be a little more detailed.

```suggestion
      terminal.write(
        """
        --destination option is no longer needed when using the latest SwiftWasm toolchains. \
        This option no longer has any effect and will be removed in a future version of `carton`. \
        You should be able to link with Foundation/XCTest without passing this option. If it is \
        still required in your build process for some reason, please report it as a bug at \
        https://github.com/swiftwasm/swift/issues/\n
        """,
        inColor: .red
      )
```

Sources/SwiftToolchain/Toolchain.swift
```diff
@@ -226,9 +221,9 @@ public final class Toolchain {
 
     terminal.write("\nBuilding the project before spinning up a server...\n", inColor: .yellow)
 
-    let builderArguments = try [
+    let builderArguments = [
       swiftPath.pathString, "build", "-c", isRelease ? "release" : "debug", "--product", product,
-      "--enable-test-discovery", "--destination", destination ?? inferDestinationPath().pathString,
+      "--enable-test-discovery", "--triple", "wasm32-unknown-wasi", "-Xswiftc", "-static-stdlib",
```
MaxDesiatov, line 226: QQ: is `-static-stdlib` now required in the latest snapshots? Should we patch SwiftPM so that it passes it to the driver automatically without requiring stand-alone SwiftPM users and `carton` to do it?

kateinoigakukun: Good question. I'm thinking about whether we should make static-stdlib to be the default for wasm target.

MaxDesiatov: I think we should, it essentially was enabled by default all this time, at least in the outcome.😄 And can you even build anything useful without linking stdlib statically?

kateinoigakukun: ok, I'll tackle this as my next task. Anyway, can we merge this pr without making the option default at now ?

MaxDesiatov (approved):

Sources/SwiftToolchain/Toolchain.swift
MaxDesiatov: Absolutely, thanks again for all this work!

TestApp/Sources/TestApp/main.swift
MaxDesiatov: Reported separately as https://github.com/swiftwasm/swift/issues/2110, feel free to update it if I got anything wrong.

carson-katri (approved):

MaxDesiatov:
@kateinoigakukun would you mind if this is merged? I want to bump the toolchain again to `wasm-5.3-SNAPSHOT-2020-10-27-a` in a separate PR to test `.wasm` extensions on binaries. Would be great to test it all end-to-end together with this PR merged in `main`, and would also avoid conflicts.

PR #142, opened 2020-10-24
Fix error message output

carson-katri:
Previously, an error message such as:
```
 ERROR   invalid redeclaration of 'resolve(in:)'
  58 |    public func resolve(in environment: EnvironmentValues) -> _Font {
```
would be output:
```
 ERROR   invalid redeclaration of 'resolve(in
  58 |    public func resolve(in environment: EnvironmentValues) -> _Font {
```
if the message had a colon.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -18,8 +18,8 @@ jobs:
           swift test -c release
           swift build -c release
           brew bundle
-          cd TestApp && ../.build/debug/carton test
-          ../.build/debug/carton bundle
+          cd TestApp && ../.build/release/carton test
+          ../.build/release/carton bundle
         # the token is required to get around GitHub API limits when downloading the toolchain
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -35,8 +35,8 @@ jobs:
           swift test -c release
           swift build -c release
           brew bundle
-          cd TestApp && ../.build/debug/carton test
-          ../.build/debug/carton bundle
+          cd TestApp && ../.build/release/carton test
+          ../.build/release/carton bundle
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
@@ -51,8 +51,8 @@ jobs:
           swift test -c release
           swift build -c release
           brew bundle
-          cd TestApp && ../.build/debug/carton test
-          ../.build/debug/carton bundle
+          cd TestApp && ../.build/release/carton test
+          ../.build/release/carton bundle
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
@@ -67,8 +67,8 @@ jobs:
           swift test -c release
           swift build -c release
           brew bundle
-          cd TestApp && ../.build/debug/carton test
-          ../.build/debug/carton bundle
+          cd TestApp && ../.build/release/carton test
+          ../.build/release/carton bundle
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
@@ -85,8 +85,8 @@ jobs:
           sudo ./install_ubuntu_deps.sh
           curl https://get.wasmer.io -sSfL | sh
           source /home/runner/.wasmer/wasmer.sh
-          cd TestApp && ../.build/debug/carton test
-          ../.build/debug/carton bundle
+          cd TestApp && ../.build/release/carton test
+          ../.build/release/carton bundle
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
@@ -103,8 +103,8 @@ jobs:
           sudo ./install_ubuntu_deps.sh
           curl https://get.wasmer.io -sSfL | sh
           source /home/runner/.wasmer/wasmer.sh
-          cd TestApp && ../.build/debug/carton test
-          ../.build/debug/carton bundle
+          cd TestApp && ../.build/release/carton test
+          ../.build/release/carton bundle
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Build on macOS 10.15 with Swift 5.2
         run: |
           sudo xcode-select --switch /Applications/Xcode_11.7.app/Contents/Developer
-          swift test -c release
+          swift test -c release --enable-test-discovery
           swift build -c release
           brew bundle
           cd TestApp && ../.build/release/carton test
@@ -32,7 +32,7 @@ jobs:
       - name: Build on macOS 11.0 with Swift 5.2
         run: |
           sudo xcode-select --switch /Applications/Xcode_11.7.app/Contents/Developer
-          swift test -c release
+          swift test -c release --enable-test-discovery
           swift build -c release
           brew bundle
           cd TestApp && ../.build/release/carton test
@@ -48,7 +48,7 @@ jobs:
       - name: Build on macOS 10.15 with Swift 5.3
         run: |
           sudo xcode-select --switch /Applications/Xcode_12.app/Contents/Developer
-          swift test -c release
+          swift test -c release --enable-test-discovery
           swift build -c release
           brew bundle
           cd TestApp && ../.build/release/carton test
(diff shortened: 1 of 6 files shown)


Package.swift
```diff
@@ -77,7 +77,11 @@ let package = Package(
     ),
     .testTarget(
       name: "CartonTests",
```
MaxDesiatov, line 79: I don't think `swift test` is actually invoked on CU, is it?

MaxDesiatov (approved):
Many thanks for the test!

Package.swift
carson-katri: There's a "Build and test" workflow but it never actually runs `swift test`, just `carton test` on the test app.

MaxDesiatov: Maybe `swift build` in it could be replaced with this?

```
swift test
swift build -c release
```

Using a release build is closer to the real environment I guess. Not sure if `swift test` should use `-c release` too?

carson-katri: I'm not sure how big a difference it will make but since release config is closer to the real environment then we should go with that.

Do you want me to add that command in this PR?

MaxDesiatov: Please do, it would verify that tests added in this PR run properly on CI.

carson-katri:
The `testExample` didn't seem to really test anything, I changed it to check the version number, but we might want to remove it?

.github/workflows/swift.yml
```diff
@@ -76,7 +80,8 @@ jobs:
 
       - name: Build on Ubuntu 18.04 with Swift 5.3
         run: |
-          swift build
+          swift test -c release
```
MaxDesiatov, line 83: **suggestion:** we won't have to maintain the test cases list manually then

```suggestion
          swift test -c release --enable-test-discovery
```

.github/workflows/swift.yml
```diff
@@ -93,7 +98,8 @@ jobs:
 
       - name: Build on Ubuntu 20.04 with Swift 5.3
         run: |
-          swift build
+          swift test -c release
```
MaxDesiatov, line 101:
```suggestion
          swift test -c release --enable-test-discovery
```

Tests/CartonTests/CartonTests.swift
```diff
@@ -1,7 +1,21 @@
...
+    #if os(macOS)
+    for bundle in Bundle.allBundles where bundle.bundlePath.hasSuffix(".xctest") {
+      return bundle.bundleURL.deletingLastPathComponent()
+    }
+    fatalError("couldn't find the products directory")
+    #else
+    return Bundle.main.bundleURL
+    #endif
+  }
+
```
MaxDesiatov, line 18: **nitpick:** please apply `swiftformat .` or run `pre-commit install` to run it automatically before every commit to avoid these linter formatting warnings.

Tests/CartonTests/CartonTests.swift
```diff
@@ -20,28 +34,68 @@ final class CartonTests: XCTestCase {
...
+    let writer = InteractiveWriter(stream: stream)
+    DiagnosticsParser().parse(testDiagnostics, writer)
+    XCTAssertEqual(stream.currentOutput, expectedOutput)
   }
 
   static var allTests = [
-    ("testExample", testExample),
+//    ("testExample", testExample),
+    ("testDiagnosticsParser", testDiagnosticsParser),
   ]
```
MaxDesiatov, line 100: These won't be needed if `--enable-test-discovery` suggestion above is applied.

Tests/CartonTests/CartonTests.swift
```diff
@@ -1,7 +1,21 @@
...
+    for bundle in Bundle.allBundles where bundle.bundlePath.hasSuffix(".xctest") {
+      return bundle.bundleURL.deletingLastPathComponent()
+    }
+    fatalError("couldn't find the products directory")
+    #else
+    return Bundle.main.bundleURL
+    #endif
+  }
+  
   func testExample() throws {
```
MaxDesiatov, line 19: I think this test is a good example of end-to-end testing for command-line invocations, maybe let's rename it and keep it?

```suggestion
  func testVersion() throws {
```

.github/workflows/swift.yml
```diff
@@ -15,7 +15,8 @@ jobs:
       - name: Build on macOS 10.15 with Swift 5.2
         run: |
           sudo xcode-select --switch /Applications/Xcode_11.7.app/Contents/Developer
-          swift build
+          swift test -c release
+          swift build -c release
           brew bundle
           cd TestApp && ../.build/debug/carton test
```
MaxDesiatov, line 21:
```suggestion
          cd TestApp && ../.build/release/carton test
```

.github/workflows/swift.yml
```diff
@@ -15,7 +15,8 @@ jobs:
...
           ../.build/debug/carton bundle
```
MaxDesiatov, line 22:
```suggestion
          ../.build/release/carton bundle
```

.github/workflows/swift.yml
```diff
@@ -31,7 +32,8 @@ jobs:
       - name: Build on macOS 11.0 with Swift 5.2
         run: |
           sudo xcode-select --switch /Applications/Xcode_11.7.app/Contents/Developer
-          swift build
+          swift test -c release
+          swift build -c release
           brew bundle
           cd TestApp && ../.build/debug/carton test
           ../.build/debug/carton bundle
```
MaxDesiatov, line 39:
```suggestion
          cd TestApp && ../.build/release/carton test
          ../.build/release/carton bundle
```

.github/workflows/swift.yml
```diff
@@ -46,7 +48,8 @@ jobs:
       - name: Build on macOS 10.15 with Swift 5.3
         run: |
           sudo xcode-select --switch /Applications/Xcode_12.app/Contents/Developer
-          swift build
+          swift test -c release
+          swift build -c release
           brew bundle
           cd TestApp && ../.build/debug/carton test
           ../.build/debug/carton bundle
```
MaxDesiatov, line 55:
```suggestion
          cd TestApp && ../.build/release/carton test
          ../.build/release/carton bundle
```

.github/workflows/swift.yml
```diff
@@ -61,7 +64,8 @@ jobs:
       - name: Build on macOS 11.0 with Swift 5.3
         run: |
           sudo xcode-select --switch /Applications/Xcode_12.1.app/Contents/Developer
-          swift build
+          swift test -c release
+          swift build -c release
           brew bundle
           cd TestApp && ../.build/debug/carton test
           ../.build/debug/carton bundle
```
MaxDesiatov, line 71:
```suggestion
          cd TestApp && ../.build/release/carton test
          ../.build/release/carton bundle
```

.github/workflows/swift.yml
```diff
@@ -76,7 +80,8 @@ jobs:
 
       - name: Build on Ubuntu 18.04 with Swift 5.3
         run: |
-          swift build
+          swift test -c release
+          swift build -c release
           sudo ./install_ubuntu_deps.sh
           curl https://get.wasmer.io -sSfL | sh
           source /home/runner/.wasmer/wasmer.sh
```
MaxDesiatov, line 87: two lines below need to be changed to use `.build/release` to fix the path

.github/workflows/swift.yml
```diff
@@ -93,7 +98,8 @@ jobs:
 
       - name: Build on Ubuntu 20.04 with Swift 5.3
         run: |
-          swift build
+          swift test -c release
+          swift build -c release
           sudo ./install_ubuntu_deps.sh
           curl https://get.wasmer.io -sSfL | sh
           source /home/runner/.wasmer/wasmer.sh
```
MaxDesiatov, line 105: `.build/debug` needs to be changed below to `.build/release`, I can't add suggestions there unfortunately

MaxDesiatov (changes requested):
`XCTestManifests.swift` and `LinuxMain.swift` also need to be deleted for test discovery to work.

kateinoigakukun:
I think we don't need to remove LinuxMain.swift. It would be helpful if following swiftpm way https://github.com/apple/swift-package-manager/blob/main/Tests/LinuxMain.swift

MaxDesiatov:
Great suggestion, I agree, keeping it with `fatalError` is a better option here!

MaxDesiatov (approved):
Awesome, thanks a lot 🙌

PR #148, opened 2020-10-30
Fix handling of binaries with `.wasm` extension

MaxDesiatov:
Latest toolchains produce binaries with the `.wasm` extension which `carton` can't handle yet.


kateinoigakukun (approved):

kateinoigakukun:
I searched a way to get produced product binary path in SwiftPM command interface, but I couldn't find it :cry:

MaxDesiatov:
Yeah, not great there isn't one 😕

PR #154, opened 2020-11-06
Use ghcr.io/swiftwasm/swift as base image

kateinoigakukun:
Update Dockerfile to use ghcr.io/swiftwasm/swift as base image and update default toolchain version as 5.3 release


MaxDesiatov:
Looks great! Let's switch to the 5.3.0 toolchain here and in the `swiftwasm-docker` repo as soon as it's tagged and assets are available.

MaxDesiatov (approved):
👍

MaxDesiatov (changes requested):
Please also update `DefaultToolchain.swift`

kateinoigakukun:
> Please also update DefaultToolchain.swift

It's already updated

MaxDesiatov (approved):
Ah, sorry I missed that 🤦‍♂️

PR #156, opened 2020-11-06
Remove wabt dependency

kateinoigakukun:
Currently, carton depends on `wabt` only to use `wasm-strip`. But wasm-strip is a really simple program, so I re-implemented it in WasmTransformer to remove wabt dependency from carton.

What do you think about this removal?


MaxDesiatov (approved):
Wonderful, this simplifies installation for us! 🙌 Removing dependencies is always great if it's easy 😄

MaxDesiatov:
This change should also shrink our Docker image by a bit, as I imagine your new transformation code is much smaller than the whole WABT package 👏

Issue #26, opened 2020-07-05
Backtraces demangling


MaxDesiatov:
In absence of proper debugging support in browsers it's very hard to debug things where there's no backtrace whatsoever, even for array out of bounds errors.

We need to investigate whether supporting [`swift-backtrace`](https://github.com/swift-server/swift-backtrace/) is viable.

MaxDesiatov:
`swift-backtrace` won't work, neither WebAssembly nor WASI support the required low-level code to my knowledge. The host environment is supposed to provide the backtraces, and it actually [does so in browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/RuntimeError). Thus, the question is more about demangling the backtraces. I don't think there's an existing Swift demangling implementation in JavaScript to use in the `dev.js` entrypoint, and it seems not that easy to pull the actual runtime implementation from the [`swiftDemangling` library](https://github.com/apple/swift/tree/master/lib/Demangling). A stopgap solution could be the [CwlDemangle](https://github.com/mattgallagher/CwlDemangle) package, but seems like that requires `Foundation`.

kateinoigakukun:
I think you can use `swift_demangle` function defined in stdlib as same as wasminspect does. https://github.com/kateinoigakukun/wasminspect/blob/master/crates/swift-runtime/src/lib.rs

MaxDesiatov:
Yes, thanks for pointing me to it. I'm also thinking we don't necessarily have to demangle in the browser, an error with its backtrace can be sent to `carton` via the WebSocket connection that's already there for hot reloads, then `carton` can print the demangled backtrace in its own output. Also much more convenient for jumping to files and lines mentioned in the backtrace if you have `carton` open in your IDE console.

MaxDesiatov:
This was fully resolved in #186.

PR #162, opened 2020-11-08
Demangle and print Firefox stack traces in terminal

MaxDesiatov:
This currently supports only Firefox, parsing of stack traces for other browsers will be added in separate PRs.

Resolves a part of #26.

It currently looks like this:

Screenshot 2020-11-08 at 19 09 03


Sources/carton/Server/Server.swift
```diff
@@ -64,8 +94,31 @@ final class Server {
...
+          switch event {
+          case let .stackTrace(rawStackTrace):
+            guard environment == .firefox else { break }
+
+            let stackTrace = rawStackTrace.firefoxStackTrace
+
+            terminal.write("\nAn error occurred, here's a stack trace for it:\n", inColor: .red)
+            stackTrace.forEach { item in
+              terminal.write("  \(item.symbol)", inColor: .cyan)
+              terminal.write(" at \(item.location)\n", inColor: .grey)
```
yonihemi, line 115: Could we consider a color other than grey? Not sure if the problem is in my terminal settings, but:
Screen Shot 2020-11-26 at 17 14 34

yonihemi (dismissed):

Sources/carton/Server/Server.swift
MaxDesiatov: Thanks for testing it, I changed it to use the default color

yonihemi (approved):

PR #165, opened 2020-11-09
Add minimum deployment target in template

kateinoigakukun:


MaxDesiatov (commented):
I've released Tokamak 0.5.1, so this could bump the version in the template too.

MaxDesiatov (approved):
Thanks!

PR #122, opened 2020-10-06
Pretty print diagnostics

carson-katri:
The output from the Swift compiler can be pretty verbose. This adds better formatting and syntax highlighting to the compiler's output (modeled after the error messages of React).

Screen Shot 2020-10-06 at 3 56 56 PM

I also modified the screen-clearing to reset the scrollback so you can easily find the first error without scrolling past it.

PR #167, opened 2020-11-11
[WIP] Align pretty-printed diagnostic messages with source code that uses horizontal whitespace characters other than space (U+0020)

WowbaggersLiquidLunch:
I noticed a few problems with the pretty-printed diagnostics. 

Given a short Tokamak program as follows:

```swift
import TokamakShim

struct TokamakApp: App {
	var body: some Scene {
		WindowGroup("Tokamak App") {
			ContentView()
		}
	}
}

struct ContentView: View {
	var body: some View {
		VStack {
			Button("foo", action: nil)
			Button("字符串", action: nil)
		}
	}
}

Ni!

TokamakApp.main()
```

`carton dev` provides the following diagnosis:

Screen Shot 2020-11-11 at 13 56 16

The most noticeable problem in the diagnostics is that the "^"s don't point to where the errors actually are. This pull request intends to address this problem by first stripping all leading whitespace in the pretty-printed source code, then copying any remaining whitespace verbatim to the indentation before each "^".

Some other problems include duplicated error messages and out-of-order diagnosis. I intend to address them in follow-up PRs.

### DIFF
--- a/Sources/CartonHelpers/Parsers/DiagnosticsParser.swift
+++ b/Sources/CartonHelpers/Parsers/DiagnosticsParser.swift
@@ -228,10 +228,11 @@ public struct DiagnosticsParser: ProcessOutputParser {
     // Output the code for this line, syntax highlighted
     let highlightedCode = Self.highlighter.highlight(message.code)
     terminal.write("  \("\(paddedLine) | ", color: "[36m")\(highlightedCode)\n") // 36: cyan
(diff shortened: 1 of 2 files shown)


WowbaggersLiquidLunch:
There are still some test cases to be added.

WowbaggersLiquidLunch:
There isn't much documentation to go by in the source. @carson-katri, since you are the author of #122, it will be great if you could take a look at the changes and see if things are done correctly.

carson-katri (commented):
LGTM except for the few linter errors, and the placeholders in the `flush` function's doc comment. Thanks for the fix!

PR #168, opened 2020-11-11
Ignore Package.resolved

WowbaggersLiquidLunch:
This PR removes and ignores `Package.resolved` from the repository. `Package.resolved` is automatically generated, and adds noise to the Git history. Additionally, with the file tracked, carton's dependencies have to be updated manually every time before it's built, and when/after `Package.swift` changes. With the file removed, `Package.swift` can serve as a single source of truth for all SwiftPM-managed dependencies.

`package-lock.json` probably should be removed and ignored as well, but I'm not entirely sure.


MaxDesiatov:
Thanks for the PR! I'd rather keep `Package.resolved`, as it guarantees that people installing it via Homebrew or on Linux from sources get the exact same list of dependencies that the developers of `carton` use to test things on their machines. Same for `package-lock.json`. Hypothetically, if any of our dependencies accidentally publish something breaking (which is especially possible as some of them are on 0.x versions, which aren't guaranteed to stricitly conform to semver), it won't unexpectedly break things for our users. 

We then can discover breakages on our side by updating `Package.resolved`, submitting it as a PR, and re-testing it on CI, which is basically what our [`update-deps.yml`](https://github.com/swiftwasm/carton/blob/main/.github/workflows/update-deps.yml) workflow does automatically every week when any updates are available. And we also make sure that our CI uses the same dependencies too in every other PR. 

It may add some noise to commit logs, but not very significant, especially as our dependencies don't have that many updates. At least safety from unexpected breakages outweighs the noise in commit logs I think.

@swiftwasm/carton-team WDYT?

WowbaggersLiquidLunch:
> it guarantees that people installing it via Homebrew or on Linux from sources get the exact same list of dependencies that the developers of `carton` use to test things on their machines. Same for `package-lock.json`.

This is a good point. I didn't think of this before.

I wonder if there is a way to have GitHub generate `Package.resolved` without having Git track it?  This way users who build from source can still get the exact same dependencies for each carton release, while pulling/cloning from GitHub won't get the `Package.resolved`. Maybe it's not worth it, and I'm thinking too much into it.

WowbaggersLiquidLunch:
Closing this PR because it doesn't seem like the right direction.

PR #186, opened 2020-12-06
Add support for Chrome and Safari stack traces

j-f1:
I made `location` optional because Safari doesn’t report it for some reason. There’s a bit of duplication between the stack trace parsers but I don’t think that’s a big problem.

I’ve also changed the behavior to log when an error occurs in an unsupported browser, and to provide a sample stack trace that we can attempt to parse.

### DIFF
--- a/Tests/CartonTests/StackTraceTests.swift
+++ b/Tests/CartonTests/StackTraceTests.swift
@@ -48,111 +48,414 @@ final class StackTraceTests: XCTestCase {
     swjs_create_function/func_ref<@webpack:///./node_modules/javascript-kit-swift/Runtime/lib/index.js?:280:28
     """.firefoxStackTrace
 
-    XCTAssertEqual(
-      stackTrace,
+    let expected: [StackTraceItem] = [
+      .init(
+        symbol: "wasmFs.fs.writeSync",
+        location: "./entrypoint/dev.js?:35:21",
+        kind: .javaScript
+      ),
+      .init(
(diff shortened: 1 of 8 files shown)


Tests/CartonTests/StackTraceTests.swift
```diff
@@ -48,111 +49,416 @@ final class StackTraceTests: XCTestCase {
...
+//          kind: .webAssembly
+//        ), .init(
+//          symbol: "swift_reportError",
+//          location: "wasm-function[21654]:0x37c242",
+//          kind: .webAssembly
+//        ), .init(
+//          symbol: "_swift_stdlib_reportFatalErrorInFile",
+//          location: "wasm-function[22950]:0x3e2996",
+//          kind: .webAssembly
+//        ),
```
MaxDesiatov, line 223: Can these comments be removed, or would you like to keep them for some purpose?

MaxDesiatov (commented):
Great work! I only have a question about commented code in tests.

Tests/CartonTests/StackTraceTests.swift
j-f1: I can remove them; I just kept them because they were in the other stack traces but not Safari’s.

MaxDesiatov: No problem, a code comment that explains why it was kept there could also work, but doesn't matter anymore 🙂

MaxDesiatov (approved):
Thank you 👍

Issue #175, opened 2020-11-23
Support resources for package dependencies


MaxDesiatov:
Currently package resources are only correctly handled for targets within the root package. Resources in packages that the root package depends on aren't handled, and even for their transitive dependencies to be handled we need to parse those packages from the `Package.resolved` file. This is probably blocked by https://github.com/swiftwasm/carton/issues/120, as I guess we need to use libSwiftPM to get that working.

MaxDesiatov:
Potentially, this would allow bundling JavaScriptKit runtime bits in the SwiftPM package instead of embedding it in `static.zip` entrypoint archives.

Issue #155, opened 2020-11-06
Sync JavaScriptKit runtime library with Package.resolved


kateinoigakukun:
AFAIU, carton supports only one version of JavaScriptKit runtime library now. If carton can dynamically determine the JavaScriptKit runtime library version based on Package.resolved and embed it, it would be really helpful.

MaxDesiatov:
I guess this is somewhat related to #120. If we use libSwiftPM for package models, we'd also get `Package.resolved` parsing capabilities too...

MaxDesiatov:
This should probably be superseded by https://github.com/swiftwasm/carton/issues/175, when JavaScriptKit runtime is shipped as a SwiftPM resource.

PR #194, opened 2020-12-25
Use libSwiftPM instead of custom model types 

### LINKED ISSUE #120: Use libSwiftPM instead of custom model types
Instead of maintaining our own packaga manifest model types, we should consider using [the SwiftPackageManifest package](https://github.com/SwiftDocOrg/SwiftPackageManifest) to offload this burden. The package is going to be reused in [`swift-doc`](https://github.com/SwiftDocOrg/swift-doc) in the future, so this feels safe to rely on.

yonihemi:
Resolves #120.

Notes:
- SwiftPM hasn't tagged new releases since March (oh the irony). I chose to point to `main` branch (and so TSC also has to point to `main` to resolve correctly) but if @swiftwasm/carton-team thinks it might be risky we can point to `release/5.3` (TSC as well) instead. I prefer main so we don't get stuck if swiftwasm advances in some 5.4 direction and we need the newer resolution.
- Names changed to reflect the distinction as [libSwiftPM does](https://github.com/apple/swift-package-manager/blob/release/5.3/Sources/PackageModel/Package.swift#L19) - `Manifest` is the PackageDescription specified in Package.swift vs `PackageModel.Package` which includes resolved dependencies. We'll use this when tackling #175 and #155.


Package.swift
```diff
@@ -26,9 +26,14 @@ let package = Package(
...
     ),
+    .package(
+      name: "SwiftPM",
+      url: "https://github.com/apple/swift-package-manager.git",
+      .branch("main")
+    ),
     .package(
       url: "https://github.com/apple/swift-tools-support-core.git",
-      .upToNextMinor(from: "0.1.10")
+      .branch("main")
```
MaxDesiatov, line 36: SwiftPM repo has tags like `swift-release-5.3.1` as far as I remember. Maybe we could rely on those?

MaxDesiatov: Disregard that, I tried with `swift-5.3.2-RELEASE` and it doesn't work because it tries to pull the `master` branch of `swift-llbuild` which doesn't exist anymore... 😕

Package.swift
```diff
@@ -76,7 +80,7 @@ let package = Package(
       name: "SwiftToolchain",
       dependencies: [
         .product(name: "AsyncHTTPClient", package: "async-http-client"),
-        .product(name: "SwiftToolsSupport-auto", package: "swift-tools-support-core"),
```
MaxDesiatov, line 79: This product dependency needs to be preserved, otherwise it causes linking issues on CI.

yonihemi: I think it's a difference in resolution between 5.2 and 5.3. Without removing these lines it doesn't resolve on 5.3, and the 5.3 test fails for other reasons.

Package.swift
yonihemi: Yes, saw that. Also the 5.3.x tags seem especially out of sync vs the `release/5.3` branches. I've opened a small [PR](https://github.com/apple/swift-package-manager/pull/3145) for the llbuild dependency.

Package.swift
MaxDesiatov: Fair enough, should be a big problem after https://github.com/swiftwasm/carton/pull/195 is merged, where we have to drop support for 5.2.

Package@swift-5.2.swift
```diff
@@ -20,9 +20,14 @@ let package = Package(
       url: "https://github.com/apple/swift-argument-parser.git",
       .upToNextMinor(from: "0.3.0")
     ),
+    .package(
+      name: "SwiftPM",
+      url: "https://github.com/apple/swift-package-manager.git",
+      .branch("main")
```
MaxDesiatov, line 26: They've just created the 5.4 branch yesterday, maybe it's worth relying on that one now instead of 5.3? Usually, when a release branch is created, they don't add much breaking changes to it if any, so probably worth trying it out.

```suggestion
      .branch("release/5.4")
```

Package@swift-5.2.swift
```diff
@@ -20,9 +20,14 @@ let package = Package(
...
+    ),
     .package(
       url: "https://github.com/apple/swift-tools-support-core.git",
-      .upToNextMinor(from: "0.1.10")
+      .branch("main")
```
MaxDesiatov, line 30:
```suggestion
      .branch("release/5.4")
```

Package@swift-5.2.swift
MaxDesiatov: Especially, as this PR already uses `main`, `release/5.4` shouldn't differ much from it.

MaxDesiatov:
@yonihemi would you mind if I pick this up?

MaxDesiatov:
Could you merge `main` again please? I've just merged the CI fix there.

Package@swift-5.2.swift
```diff
@@ -80,7 +85,6 @@ let package = Package(
       name: "CartonHelpers",
       dependencies: [
         .product(name: "AsyncHTTPClient", package: "async-http-client"),
-        .product(name: "SwiftToolsSupport-auto", package: "swift-tools-support-core"),
```
MaxDesiatov, line 83: This needs to stay to avoid linker errors.

Package.swift
yonihemi: I don't have access to develop on 5.2 ATM and making guess PRs until it `works` is not great. I suggest we merge #195 and move on.

MaxDesiatov: Uh, I forgot about this bit. I can test with 5.2 on my side.

yonihemi: I mean moving forward, since our dependency graph is large and there are (undocumented?) differences between the versions, plus 5.3 is stable and we have 5.4 to worry about - it's a good time to cut the chord IMO.

MaxDesiatov: Sure, it's the lack of output redirection with WAMR that I'm worried about. I'll try to get that sorted soon and we'll merge #195 after that.

MaxDesiatov (approved):
Today is the day! 👏

MaxDesiatov:
@yonihemi ready to merge 🎉

PR #195, opened 2020-12-25
Use WAMR for `carton test`

### LINKED ISSUE #177: Make Wasmer an optional dependency?
After https://github.com/swiftwasm/carton/issues/42 is resolved, we'll be able to run tests in browsers. We should still allow using Wasmer, but I think it's easier to expect that people have a browser installed than Wasmer? Especially as currently it's not as easy to install Wasmer on Apple Silicon via Homebrew. WDYT?

MaxDesiatov:
This simplifies our CI scripts, and installation via Homebrew and on Linux. It also allows us to pin the version of WASI runtime we use for CLI test execution.

Closes #177.


yonihemi (approved):

.github/workflows/swift.yml
```diff
@@ -7,51 +7,16 @@ on:
...
       - name: Build on macOS 10.15 with Swift 5.3
         run: |
-          sudo xcode-select --switch /Applications/Xcode_12.app/Contents/Developer
-          swift test -c release --enable-test-discovery
+          sudo xcode-select --switch /Applications/Xcode_12.3.app/Contents/Developer
           swift build -c release
+          swift test -c release --enable-test-discovery
           brew bundle
           cd TestApp && ../.build/release/carton test
           ../.build/release/carton test --environment defaultBrowser
```
thecb4, line 22: This can be removed. It's included in the CI Tests now.

If we are going to build before tests it should be

```
swift build --build-tests --enable-test-discovery
swift test --enable-test-discovery
```

.github/workflows/swift.yml
```diff
@@ -66,9 +31,9 @@ jobs:
       - uses: actions/checkout@v2
       - name: Build on macOS 11.0 with Swift 5.3
         run: |
-          sudo xcode-select --switch /Applications/Xcode_12.2.app/Contents/Developer
-          swift test -c release --enable-test-discovery
+          sudo xcode-select --switch /Applications/Xcode_12.3.app/Contents/Developer
           swift build -c release
+          swift test -c release --enable-test-discovery
```
thecb4, line 36: Similar to last comment. CI takes care of `carton test` now.

Issue #99, opened 2020-08-28
Test all commands on CI


MaxDesiatov:
None of the implemented commands are currently tested end-to-end.

thecb4:
Done:
- [x] bundle (no arguments)
- [x] dev (no arguments)
- [x] init (no arguments)

SDK
- [ ] install subcommand
- [ ] version subcommand
- [ ] local subcommand

Init
- [ ] with template option
- [ ] list-templates subcommand

Bundle
- [ ] with executable option
- [ ] with custom index option

Dev
- [ ] with custom index option
- [ ] with release option
- [ ] with verbose option
- [ ] with port option

Package
- [ ] with no arguments
- [ ] with any arguments

Test
- [ ] with no arguments
- [ ] with release option
- [ ] list option
- [ ] test-cases option

PR #198, opened 2020-12-29
Add `test` command test with no arguments

thecb4:
This is in support of #99 

Added
* Test Command Test with no arguments

Changed
* Moved TestApp, entrypoint, package.json, and package-lock.json folder to Test/Fixtures/carton-test
* Script for GitHub Actions now builds (including tests) then runs the tests separately. 
* Dockerfile now builds (including tests) then runs the tests separately. 

Removed
* Script for GitHub Actions no longer references TestApp
* Dockerfile no longer references TestApp


MaxDesiatov (changes requested):
Could you move the `entrypoint` directory and `package.json`/`package-lock.json` back to the root from the `Fixtures` dir? It's used for entrypoints for all commands, not just in the fixture.

thecb4:
Sure - moved them back.

MaxDesiatov:
And the `entrypoint` directory too please 🙂

MaxDesiatov (approved):
LGTM, thank you 👏

PR #202, opened 2021-01-01
Fix browser testing for Safari, update tasks.json

### LINKED ISSUE #200: `carton test --environment defaultBrowser` broken on GitHub Actions
While the process itself seems to run and return 0 status code, it doesn't look like any of the tests have actually run. We need to verify that a browser instance is actually launched and check why no tests appeear in the report if it does launch.

MaxDesiatov:
Resolves #200 for macOS. We still probably need to prepare an image (maybe separate image to keep the primary image slim enough) with pre-installed browsers for testing on Linux to work.

This should fix GitHub Actions not running any of the browser tests on macOS. Thanks to @thecb4 for finding the cause of the issue!

I've also updated VSCode's `tasks.json` with new fixtures paths and simplified subdirectory nesting for `TestApp`.


thecb4 (commented):
Unclear why the comment code was added back in. I know the team tries to keep things clean.

MaxDesiatov:
Are you referring to the commented code removed in https://github.com/swiftwasm/carton/pull/202/commits/9f669f22926d2d881c2eb721de340dddfae5f530? Sorry about that, just some experiments I've done locally slipped in. They are removedin the latest commit here.

yonihemi (approved):

PR #204, opened 2021-01-01
Avoid running tests while building the Docker image

MaxDesiatov:
We already run these tests on Ubuntu as a part of separate CI jobs.


thecb4 (approved):
Looks good to me. Assuming we will remove the --enable-test-discovery at a later point.

PR #206, opened 2021-01-06
[NFC] Remove `Project.inPlace` to simplify package init logic

WowbaggersLiquidLunch:
`Project.inPlace` is used only once, when initialising a new package through SwiftPM. And it's somewhat redundant along with `.name` when `.path` is really the only thing needed.

The logic can be simplified even further if/when we use libSwiftPM directly.

This PR contains no functional change, but is API-breaking for any other project that depends on CartonKit or SwiftToolchain.

### DIFF
--- a/Sources/CartonCLI/Commands/Init.swift
+++ b/Sources/CartonCLI/Commands/Init.swift
@@ -25,43 +25,85 @@ struct Init: ParsableCommand {
     subcommands: [ListTemplates.self]
   )
 
-  @Option(name: .long,
-          help: "The template to base the project on.",
-          transform: { Templates(rawValue: $0.lowercased()) })
+  @Option(
+    name: .long,
+    help: "The template to base the project on.",
+    transform: { Templates(rawValue: $0.lowercased()) }
+  )
   var template: Templates?
 
-  @Option(name: .long,
-          help: "The name of the project") var name: String?
+  @Option(
+    name: .long,
+    help: """
+          The name of the project, if it's different from the current working directory's name.
+          If this option is set, carton will create the package for the project in a subdirectory by the given name.
+          Otherwise, the package will be created in the current working directory, with the package named after the directory's.
+          """
+  )
+  var name: String?
 
   func run() throws {
     let terminal = InteractiveWriter.stdout
 
-    guard let name = name ?? localFileSystem.currentWorkingDirectory?.basename else {
-      terminal.write("Project name could not be inferred\n", inColor: .red)
-      return
+    guard let currentDirectory = localFileSystem.currentWorkingDirectory else {
+      throw ProjectInitialisationError.cannotFindCurrentDirectory
     }
-    guard let currentDir = localFileSystem.currentWorkingDirectory else {
-      terminal.write("Failed to get current working directory.\n", inColor: .red)
-      return
-    }
+
+    /// The kind of template the project is to be created with.
+    /// - Note: This value is of the type `Templates`, not `Template`.
     let template = self.template ?? .basic
+
+    let projectName = name ?? currentDirectory.basename
+    let projectPath = name == nil ? currentDirectory : currentDirectory.appending(component: projectName)
+    
+    // FIXME: Replace with `NSString.abbreviatingWithTildeInPath`?
+    let packagePathRelativeToHomeDirectory = projectPath.relative(to: localFileSystem.homeDirectory)
+
     terminal.write("Creating new project with template ")
     terminal.write("\(template.rawValue)", inColor: .green)
     terminal.write(" in ")
-    terminal.write("\(name)\n", inColor: .cyan)
+    terminal.write("~/\(packagePathRelativeToHomeDirectory.pathString)\n", inColor: .cyan)
 
-    guard let packagePath = self.name == nil ?
-      localFileSystem.currentWorkingDirectory :
-      AbsolutePath(name, relativeTo: currentDir)
-    else {
-      terminal.write("Path to project could be created.\n", inColor: .red)
-      return
+    do {
+      // `LocalFileSystem.createDirectory` wraps `FileManager.createDirectory`, which throws an NSError.
+      try localFileSystem.createDirectory(projectPath)
+    } catch let nsError as NSError {
+        throw ProjectInitialisationError.cannotCreateDirectory(path: packagePathRelativeToHomeDirectory, reason: nsError.localizedDescription)
     }
-    try localFileSystem.createDirectory(packagePath)
-    try template.template.create(
-      on: localFileSystem,
-      project: .init(name: name, path: packagePath, inPlace: self.name == nil),
-      terminal
-    )
+    
+    // FIXME: Handle this error-catching in each `Template.create()` instead.
+    do {
+      // FIXME:  The first `template` is of type `Templates`, and two consecutive `template` looks odd.
+      // Maybe rename `Templates` as `TemplateKind`, and refactor `Template` so the template doesn't have to be created this way.
+      try template.template.createProject(
+        at: projectPath,
+        inSubdirectory: name != nil,
+        on: localFileSystem,
+        terminal: terminal
+      )
+    } catch let nsError as NSError {
+        throw ProjectInitialisationError.cannotCreateWithTemplate(template, reason: nsError.localizedDescription)
+    } catch let error {
+        throw ProjectInitialisationError.cannotCreateWithTemplate(template, reason: "\(error)")
+    }
   }
+}
+
+// TODO: Add a diagnostics engine for errors thrown.
+
+enum ProjectInitialisationError: Error, CustomStringConvertible {
+    case cannotFindCurrentDirectory
+    case cannotCreateDirectory(path: RelativePath, reason: String)
+    case cannotCreateWithTemplate(Templates, reason: String)
+    
+    var description: String {
+        switch self {
+        case .cannotFindCurrentDirectory:
+            return "Cannot find the current working directory."
+        case let .cannotCreateDirectory(path, reason):
+            return "Cannot create directory ~/\(path.pathString): \(reason)"
+        case let .cannotCreateWithTemplate(template, reason):
+            return "Cannot create project with template '\(template)': \(reason)"
+        }
+    }
 }
--- a/Sources/CartonKit/Model/Template.swift
+++ b/Sources/CartonKit/Model/Template.swift
@@ -30,11 +30,12 @@ public enum Templates: String, CaseIterable {
 
 public protocol Template {
   static var description: String { get }
-  static func create(
-    on fileSystem: FileSystem,
-    project: Project,
-    _ terminal: InteractiveWriter
-  ) throws
+    static func createProject(
+      at projectPath: AbsolutePath,
+      inSubdirectory createInSubdirectory: Bool,
+      on fileSystem: FileSystem,
+      terminal: InteractiveWriter
+    ) throws
 }
 
 enum TemplateError: Error {
@@ -74,13 +75,18 @@ struct TargetDependency: CustomStringConvertible {
 
 extension Template {
   static func createPackage(
+    for project: Project,
     type: PackageType,
-    fileSystem: FileSystem,
-    project: Project,
-    _ terminal: InteractiveWriter
+    inSubdirectory createInSubdirectory: Bool,
+    on fileSystem: FileSystem,
+    terminal: InteractiveWriter
   ) throws {
     try Toolchain(fileSystem, terminal)
-      .packageInit(name: project.name, type: type, inPlace: project.inPlace)
+      .runPackageInit(
+        name: project.name,
+        type: type,
+        initialiseInSubdirectory: createInSubdirectory
+      )
   }
 
   static func createManifest(
@@ -131,13 +137,25 @@ extension Templates {
   struct Basic: Template {
     static let description: String = "A simple SwiftWasm project."
 
-    static func create(
+    static func createProject(
+      at projectPath: AbsolutePath,
+      inSubdirectory createInSubdirectory: Bool,
       on fileSystem: FileSystem,
-      project: Project,
-      _ terminal: InteractiveWriter
+      terminal: InteractiveWriter
     ) throws {
+
+      let project = Project(path: projectPath)
+
       try fileSystem.changeCurrentWorkingDirectory(to: project.path)
-      try createPackage(type: .executable, fileSystem: fileSystem, project: project, terminal)
+
+      try createPackage(
+        for: project,
+        type: .executable,
+        inSubdirectory: createInSubdirectory,
+        on: fileSystem,
+        terminal: terminal
+      )
+
       try createManifest(
         fileSystem: fileSystem,
         project: project,
@@ -153,6 +171,7 @@ extension Templates {
         ],
         terminal
       )
+
     }
   }
 }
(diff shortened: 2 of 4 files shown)


Sources/SwiftToolchain/Toolchain.swift
```diff
@@ -283,21 +283,21 @@ public final class Toolchain {
...
-    }
-    try ProcessRunner(initArgs, terminal)
-      .waitUntilFinished()
+	
+	if initialiseInSubdirectory {
+		arguments += ["--name", name]
+	}
+	
+    try ProcessRunner(arguments, terminal).waitUntilFinished()
   }
```
WowbaggersLiquidLunch, line 297: This function can be made simpler if we replace it with something like this:

```swift
public func initialisePackage(type: PackageType, for project: Project) throws {
    try localFileSystem.createDirectory(project.path) // should already be created in `Init.run()`
    guard let currentDirectory = localFileSystem.currentWorkingDirectory else {
        throw someError
    }
    let projectRelativePath = project.path.relative(to: currentDirectory)
    let arguments = [
        swiftPath.pathString, "package", "init",
        "--type", type.rawValue,
        "--name", projectRelativePath.pathString
    ]
    try ProcessRunner(arguments, terminal).waitUntilFinished()
}
```

If this function is simplified, other functions that call it can be simplified as well, without having to carry the extra `initialiseInSubdirectory` information. However, I'm not sure if this is making too much assumption around the current directory path and the project's path.

Sources/CartonCLI/Commands/Init.swift
```diff
@@ -25,43 +25,85 @@ struct Init: ParsableCommand {
...
+        switch self {
+        case .cannotFindCurrentDirectory:
+            return "Cannot find the current working directory."
+        case let .cannotCreateDirectory(path, reason):
+            return "Cannot create directory ~/\(path.pathString): \(reason)"
+        case let .cannotCreateWithTemplate(template, reason):
+            return "Cannot create project with template '\(template)': \(reason)"
+        }
+    }
+}
```
WowbaggersLiquidLunch, line 109: It seems more appropriate to me that carton should throw an error instead of simply writing a message on the terminal. But we probably need a diagnostics engine to fully take advantage of it.

Sources/CartonCLI/Commands/Init.swift
```diff
@@ -25,43 +25,85 @@ struct Init: ParsableCommand {
...
+      )
+    } catch let nsError as NSError {
+        throw ProjectInitialisationError.cannotCreateWithTemplate(template, reason: nsError.localizedDescription)
+    } catch let error {
+        throw ProjectInitialisationError.cannotCreateWithTemplate(template, reason: "\(error)")
     }
-    try localFileSystem.createDirectory(packagePath)
-    try template.template.create(
-      on: localFileSystem,
-      project: .init(name: name, path: packagePath, inPlace: self.name == nil),
```
WowbaggersLiquidLunch, line 63: I changed this function from taking a `Project` to taking a project's path as an argument. Because there is some weird error of `Project`'s synthesized member-wise initializer having `internal` access level, although `Project` is declared `public`.

WowbaggersLiquidLunch:
@carson-katri I think you are the previous author of most of the things changed in this PR. Could you take a look at them and see if everything is done properly? Thanks!

WowbaggersLiquidLunch:
Althouth the PR passed all CI tests, somehow it didn't pass some tests locally:

image

However, the test failures only appear in Xcode's report navigator, but not issue navigator. I'm not sure why this is happening.

Sources/CartonKit/Model/Template.swift
```diff
@@ -30,11 +30,12 @@ public enum Templates: String, CaseIterable {
 
 public protocol Template {
```
MaxDesiatov, line 31: While we're at it, how's about renaming this protocol to `TemplateProtocol` so that `enum Templates` could become `enum Template`? This would reduce any potential confusion.

MaxDesiatov: Another option is keeping `protocol Template` as is, but renaming `enum Template` to `enum TemplateName`.

Sources/CartonCLI/Commands/Init.swift
```diff
@@ -25,43 +25,85 @@ struct Init: ParsableCommand {
...
+    let projectPath = name == nil ? currentDirectory : currentDirectory.appending(component: projectName)
+    
+    // FIXME: Replace with `NSString.abbreviatingWithTildeInPath`?
+    let packagePathRelativeToHomeDirectory = projectPath.relative(to: localFileSystem.homeDirectory)
+
     terminal.write("Creating new project with template ")
     terminal.write("\(template.rawValue)", inColor: .green)
     terminal.write(" in ")
-    terminal.write("\(name)\n", inColor: .cyan)
+    terminal.write("~/\(packagePathRelativeToHomeDirectory.pathString)\n", inColor: .cyan)
```
MaxDesiatov, line 65: How's about displaying the full path? While the majority of projects would probably be in the home directory, as soon as one is outside this path looks weird. And the amount of characters saved by tilde is not that huge I think to warrant the hassle.

Sources/CartonCLI/Commands/Init.swift
```diff
@@ -25,43 +25,85 @@ struct Init: ParsableCommand {
...
+      // `LocalFileSystem.createDirectory` wraps `FileManager.createDirectory`, which throws an NSError.
+      try localFileSystem.createDirectory(projectPath)
+    } catch let nsError as NSError {
+        throw ProjectInitialisationError.cannotCreateDirectory(path: packagePathRelativeToHomeDirectory, reason: nsError.localizedDescription)
+    }
+    
+    // FIXME: Handle this error-catching in each `Template.create()` instead.
+    do {
+      // FIXME:  The first `template` is of type `Templates`, and two consecutive `template` looks odd.
+      // Maybe rename `Templates` as `TemplateKind`, and refactor `Template` so the template doesn't have to be created this way.
```
MaxDesiatov, line 77: I agree with the renaming, as I've mentioned in comments in the respective file. Feel free to rename in this PR or a separate one if you prefer that.

Sources/CartonCLI/Commands/Init.swift
```diff
@@ -25,43 +25,85 @@ struct Init: ParsableCommand {
...
+      try template.template.createProject(
+        at: projectPath,
+        inSubdirectory: name != nil,
+        on: localFileSystem,
+        terminal: terminal
+      )
+    } catch let nsError as NSError {
```
MaxDesiatov, line 84: Plain `Error` type has the `localizedDescription` property too, is there a need to cast to `NSError` here after all?

Sources/CartonCLI/Commands/Init.swift
MaxDesiatov: That's the default behavior in Swift for `structs` as far as I know: member-wise initializer's are always `internal` even in `public` types, to get a public one you have to declare it explicitly.

Sources/CartonCLI/Commands/Init.swift
MaxDesiatov: Yes, the reason for writing terminal messages is that thrown errors aren't nicely formatted or highlighted by default.

Sources/SwiftToolchain/Toolchain.swift
```diff
@@ -283,21 +283,21 @@ public final class Toolchain {
...
     ]
-    if !inPlace {
-      initArgs.append(contentsOf: ["--name", name])
-    }
-    try ProcessRunner(initArgs, terminal)
-      .waitUntilFinished()
+	
+	if initialiseInSubdirectory {
+		arguments += ["--name", name]
+	}
```
MaxDesiatov, line 294: Formatting here looks weird. Could you run `swiftformat` on the whole project please? If you run `pre-commit install` [as described in `README.md`](https://github.com/swiftwasm/carton#coding-style) this will install a pre-commit hook that will run SwiftLint and SwiftFormat before you commit.

Sources/SwiftToolchain/Toolchain.swift
```diff
@@ -283,21 +283,21 @@ public final class Toolchain {
     return testBundlePath
   }
 
-  public func packageInit(name: String, type: PackageType, inPlace: Bool) throws {
-    var initArgs = [
+  public func runPackageInit(name: String, type: PackageType, initialiseInSubdirectory: Bool) throws {
```
MaxDesiatov, line 286: I'm all for British spelling dominating the world, but consistency wins after all. Let's use the US one in code for consistency with other frameworks that have `Color` and other heretical spelling variants in them 😄

```suggestion
  public func runPackageInit(name: String, type: PackageType, initializeInSubdirectory: Bool) throws {
```

Sources/CartonKit/Model/Template.swift
```diff
@@ -216,6 +245,7 @@ extension Templates {
         """
         .write(to: $0)
       }
+
```
MaxDesiatov, line 248: This one's going to be cleaned up by SwiftFormat anyway I think

```suggestion
```

Sources/SwiftToolchain/Toolchain.swift
MaxDesiatov: Yeah, what if `projectRelativePath` evaluates to `.`?

MaxDesiatov (changes requested):
The general direction seems legit, I left a few comments. Could you also fix SwiftLint warnings as highlighted here in the diff? Thanks!

Sources/SwiftToolchain/Toolchain.swift
WowbaggersLiquidLunch: I experimented a bit. `.` and `..` and anything but a single undelimited path component is not going to work. 

If I do `swift package init --name a/b/../c`, then it will create a structure like this:

```
.
├── Package.swift
├── README.md
├── Sources
│   └── a
│       ├── b
│       └── c
│           └── main.swift
└── Tests
    ├── LinuxMain.swift
    └── a
        └── cTests
            ├── XCTestManifests.swift
            └── a_b____cTests.swift
```

I also found that SwiftPM always creates new projects "in place" and never creates a project in a subdirectory of the current directory. Should carton mirror what SwiftPM does?

MaxDesiatov: I hope that by delegating to `swift package init` as much as possible we would essentially mirror their behavior, right?

Sources/CartonCLI/Commands/Init.swift
WowbaggersLiquidLunch: I was thinking more in terms of privacy. Sometimes users (at least I) need to take screenshots of terminal output, but don't want to share their usernames. I can change it back to an absolute path.

MaxDesiatov: Oh, that's a good point. Maybe it should just check if a given path is a subpath of (i.e. starts with) the home path to display with a tilde, but to display the root path otherwise?

Sources/SwiftToolchain/Toolchain.swift
WowbaggersLiquidLunch: Yep, I think so. Although, by mirroring SwiftPM, carton will have to remove the ability of creating new project in a subdirectory. I don't know if this is a right thing to do.

If carton should behave like SwiftPM, then we can probably just special-case `.` for `projectRelativePath`.

Sources/CartonCLI/Commands/Init.swift
WowbaggersLiquidLunch: Do you think it would be a good idea to use `NSString.abbreviatingWithTildeInPath`? The cost should be low for just this one-time use, and the module already imports Foundation.

MaxDesiatov: Yeah, sure, why not?

Sources/CartonCLI/Commands/Init.swift
WowbaggersLiquidLunch: I tried to look for the implementation of `localizedDescription` for `Error`, and all I found is [this](https://github.com/apple/swift/blob/4fbdaf53eadea47e3dabe13f9111f5d60cab6bce/stdlib/public/Darwin/Foundation/NSError.swift#L213):

```swift
public extension Error {
  /// Retrieve the localized description for this error.
  var localizedDescription: String {
    return (self as NSError).localizedDescription
  }
}
```

I don't know if all errors can be cast as `NSError`. I suppose I can just do this:

```swift
catch let error {
  throw ProjectInitialisationError.cannotCreateWithTemplate(template, reason: error.localizedDescription)
}
```

but I don't know what would happen to errors that can't be cast as `NSError`.

I think if carton ever uses libSwiftPM directly, then we can use its diagnostics engine, and all these "catch and throw something different" won't be needed anymore.

Sources/SwiftToolchain/Toolchain.swift
WowbaggersLiquidLunch: Ok. I'll change the spellings. I can't commit the suggested change directly, though. Because I need to change the call sites as well.

Sources/SwiftToolchain/Toolchain.swift
WowbaggersLiquidLunch: I'll run `swiftformat` and `pre-commit install`. I didn't read that part of `README` previously.

MaxDesiatov: Thanks, and no problem at all 🙂

Sources/SwiftToolchain/Toolchain.swift
MaxDesiatov: Sorry, I'm a bit confused. Wouldn't the simplified version, as you've shown in `initialisePackage` in the original comment here, mirror SwiftPM behavior completely?

WowbaggersLiquidLunch: I think I might be confusing myself here as well. 

My understanding is that there is a difference between these 2 commands: 

`swift package init`

`swift package init --name .`

The first command initialises a project with using the name of the currently working directory. Whereas the second command initialises a project named "." and with "." replaced by "_" in source file names.

In carton's logic (both in trunk and in this PR), when the user sets the `--name` option in `carton init`, the project's path is recorded as `currently/working/directory/projectName`. So if the user does `carton init --name .`, then the project's path evaluates to the current directory.

As I'm writing this comment, I realise that `initialisePackage` can completely mirror SwiftPM's behaviour, if I change `Project` to include only the project's name, but not its path. And there is no need for special-casing `.`. Currently in this PR, `Project` includes only the path, and `initialisePackage` is not able to distinguish between `carton init` and `carton init --name .` from just the path.

MaxDesiatov: Right, I think that `--name .` is such a rare case that we're unlikely to see anyone stumbling upon this inconsistency in practice. But I also don't mind if you clean it up in this PR or separately, or not at all given that it's an obscure edge case.

Sources/CartonCLI/Commands/Init.swift
MaxDesiatov: This option seems legit to me:
```swift
catch let error {
  throw ProjectInitialisationError.cannotCreateWithTemplate(template, reason: error.localizedDescription)
}
```

WowbaggersLiquidLunch: I just tried this in the REPL:

```swift
enum ExampleError: Error, CustomStringConvertible { 
    case example1 
    var description: String { 
        "this is an example error" 
    }
}

let anError = ExampleError.example1
print(anError.localizedDescription)

error: repl.swift:9:15: error: value of type 'ExampleError' has no member 'localizedDescription'
print(anError.localizedDescription)
      ~~~~~~~ ^~~~~~~~~~~~~~~~~~~~
```

It gives the same error if `ExampleError` doesn't conform to `CustomStringConvertible`.

MaxDesiatov: Well, this `NSError` business gets ugly really quickly, consider this:

```swift
import Foundation

enum ExampleError: Error, CustomStringConvertible {
    case example1 
    var description: String { 
        "this is an example error" 
    }
}

do {
    throw ExampleError.example1
} catch let nsError as NSError {
    print("nserror: \(nsError.localizedDescription)")
} catch {
    print("error \(error)")
}
```

prints `nserror: The operation couldn’t be completed. (ConcurrencyPlayground.ExampleError error 0.)`. That is, any error can be cast to `NSError`, but its `localizedDescription` ignores the `description` value. The `catch let error` branch is never evaluated, because the first `NSError` branch always takes precedence no matter the actual type of the error. At least from this perspective there should be only one branch, either handling it as `NSError` or `Error`. Having two branch is redundant.

MaxDesiatov:
Hi @WowbaggersLiquidLunch, would you be interested in picking this up again in the near future?

WowbaggersLiquidLunch:
Hi @MaxDesiatov Sorry for the delay in my reply and the long time since I last worked on this.

I plan on revisiting this PR. I'm not able to do it right now, but I think I'll do it within about a month. It's one of my backlogged things that I want to pick back up over summer. 

If you prefer, I can close this PR for now, and reopen it later when I start working on it again.

kateinoigakukun:
We no longer maintain `init` subcommand because SwiftPM's `init` subcommand is already WebAssembly compatible.

PR #214, opened 2021-01-17
Update dependencies

MaxDesiatov:
- Dependency updates

Auto-generated by [create-pull-request][1]

[1]: https://github.com/peter-evans/create-pull-request


thecb4:
Would it make sense to move to using "exact" instead of "upto..." and plan regular dependency update? 

That would eliminate the need to change the Package.resolved.

I know that's hard with not everyone using Semantic Versioning.

MaxDesiatov:
With `exact` we'd have to bump versions manually even for minor bugfix updates. Say our dependency tagged an important bugfix release, we'll have to get notified about that (most probably we won't be), and then manually bump it in `Package.swift`.

Currently version bumps like this PR are fully automated with GitHub Actions. If something important breaks, such automated PRs won't pass our CI and won't be merged until the issue is resolved.

thecb4:
Ah. They are automated, ok. Automatically looks at the latest version and starts a PR?

Cheers,
CB

MaxDesiatov:
Yes. There's a cron job set up with GitHub Actions that regularly calls `swift package update` and commits the resulting update to `Package.resolved` if any is available to the `update-dependencies` branch .

PR #218, opened 2021-01-31
Add tests for `sdk versions` and `sdk local` commands

thecb4:
This is in support of #99 

SDK tests for:
[x] version
[x] local


Brewfile
```diff
@@ -3,3 +3,4 @@ brew "swiftformat"
 brew "swiftlint"
 brew "binaryen"
 brew "wasmer"
+brew "kylef/formulae/swiftenv"
```
MaxDesiatov, line 6: Is this one still needed?

thecb4: Not with the changes. I'll remove.

MaxDesiatov (approved):
Fantastic, thanks!

PR #221, opened 2021-02-14
Add `carton init` with template test. Supports #99

thecb4:
Pretty straight forward test. I am considering removing the full source comparison and only looking for "import TokamakDOM" and then including a lookup in the Package.swift file for the "tokamak repository"

### DIFF
--- a/Tests/CartonCommandTests/InitCommandTests.swift
+++ b/Tests/CartonCommandTests/InitCommandTests.swift
@@ -73,4 +73,84 @@ final class InitCommandTests: XCTestCase {
     // finally, clean up
     try packageDirectory.delete()
   }
+
+  func testInitWithTokamakTemplate() throws {
+    // given I've created a directory
+    let package = "fusion"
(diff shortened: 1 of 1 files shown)


Tests/CartonCommandTests/InitCommandTests.swift
```diff
@@ -73,4 +73,84 @@ final class InitCommandTests: XCTestCase {
     // finally, clean up
     try packageDirectory.delete()
   }
+
+  func testInitWithTokamakTemplate() throws {
+    // given I've created a directory
+    let package = "fusion"
+    let packageDirectory = testFixturesDirectory.appending(component: package)
+    let expectedTemplateSource =
```
MaxDesiatov, line 81: Could this string be moved outside of the function body? Then the linter warning about the function body line count would go away.

thecb4: It can. Easy fix.

MaxDesiatov (approved):
Awesome, thanks!

PR #236, opened 2021-05-03
Bring wasm-strip back into the container image

flavio:
It looks like the carton image (and hence the swift-action one, which is derived from it), doesn't have the `wasm-strip` binary anymore.

This PR brings back the binary into the image. While doing that, it also installs the different WebAssembly tools from Ubuntu repositories.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ RUN export DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true && ap
   libsqlite3-0 \
   libsqlite3-dev \
   curl unzip \
+  wabt binaryen \
   && export WASMER_DIR=/usr/local && curl https://get.wasmer.io -sSfL | sh && \
   rm -r /var/lib/apt/lists/*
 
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@flavio Sorry for the delayed review. Would mind clarifying how you're going to use `wasm-strip`? We've recently added a transformer pass to `carton`, which strips debug symbols, so `wasm-strip` is no longer needed for that. Having it in our scripts only makes installation and CI scripts run for longer time with no added benefit?

flavio:
I have a WASI application that I'm compiling using the [swiftwasm-action](https://github.com/swiftwasm/swiftwasm-action/).
This GitHub action image is based on the carton one, that's why I made the PR to bring back `wasm-strip`.

The main reason is: I can't use carton to build my app, and I still need to strip the Wasm file before I run wasm-opt against it.

Would you prefer the PR to be made against https://github.com/swiftwasm/swiftwasm-action/blob/main/Dockerfile

MaxDesiatov:
Yes please, the `swiftwasm-action` repository would be more appropriate for this. Thanks!

flavio:
@MaxDesiatov thanks for the guidance. I'm closing this PR in favor of https://github.com/swiftwasm/swiftwasm-action/pull/2

Issue #181, opened 2020-12-01
Use `SerializedDiagnostics` for diagnostics parsing


MaxDesiatov:
This is already implemented in [the latest stable version of TSC](https://github.com/apple/swift-tools-support-core/blob/aab25103fa74f00ee4dc651c723f05b946395037/Sources/TSCUtility/SerializedDiagnostics.swift#L13). Required output files can be emitted by passing ` -Xswiftc -serialize-diagnostics` to `swift build`. This will allow us to keep the output of `swift build` untouched during the build process, thus preserving the build progress. Build progress output is currently absent, but was available before our own diagnostic parsing from `swift build` stdout was introduced.

kateinoigakukun:
Build error diagnostics is now delegated to SwiftPM

PR #239, opened 2021-05-23
Fix `carton dev` crashing with SO sanitizer

MaxDesiatov:
Previously, the file we need to include with the build to enable the stack overflow sanitizer was written as a temporary file. This didn't happen though on watcher rebuilds with `carton dev` and caused crashes when a code line containing `try!` relying on this file to be present was executed.

I think it's easier and more efficient to bundle this file in the `static.zip` file that already includes our JS entrypoints. We require `static.zip` to be downloaded and unpacked successfully into `~/.carton/static` before every build anyway. It makes more sense to sense the sanitizer file in `~/.carton/static/so_sanitizer.wasm` and use that instead writing and deleting it at a temporary path on every build.

I've also updated the JSKit dependency and cleaned up some linter warnings by moving large tuple values into a separate `BuildDescription` type.


j-f1 (dismissed):

Sources/SwiftToolchain/Builder.swift
```diff
@@ -49,12 +49,12 @@ public final class Builder {
...
 
-  public func run() -> AnyPublisher<String, Error> {
+  private func processPublisher(builderArguments: [String]) -> AnyPublisher<String, Error> {
     let buildStarted = Date()
     let process = ProcessRunner(
-      arguments,
+      builderArguments,
       loadingMessage: "Compiling...",
-      parser: DiagnosticsParser(),
+      parser: nil,
```
MaxDesiatov, line 57: FYI @carson-katri I'm disabling this parser for now since it sometimes fails to parse project errors, but lets through the GTK warning. This causes confusing diagnostic messages as reported in https://github.com/swiftwasm/carton/issues/241. A better long-term solution would be to implement https://github.com/swiftwasm/carton/issues/181 instead, I hope to get to it at some point, unless someone else is willing to take this over.

PR #240, opened 2021-05-23
Avoid building in `release` mode when testing

MaxDesiatov:
This fixes the issue with Linux builds, as Linux CI hosts were updated to Swift 5.4 which no longer supports testing with release builds.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -24,9 +24,9 @@ jobs:
           #   swift_version: 5.3
           #   xcode: /Applications/Xcode_12.2.app/Contents/Developer
           - os: ubuntu-18.04
-            swift_version: 5.3
+            swift_version: 5.4
           - os: ubuntu-20.04
-            swift_version: 5.3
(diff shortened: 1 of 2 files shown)


.github/workflows/swift.yml
```diff
@@ -58,14 +58,14 @@ jobs:
           mkdir -p $HOME/.carton
           cp -r static $HOME/.carton/static
       - name: Build Project
-        run: swift build -c release --build-tests --enable-test-discovery
```
MaxDesiatov, line 61: Swift 5.4 marks `--enable-test-discovery` as deprecated.

Tests/LinuxMain.swift
```diff
@@ -1 +0,0 @@
-fatalError("Use `swift test --enable-test-discovery` to run tests")
```
MaxDesiatov, line 1: This file is no longer needed since test discovery is enabled by default in Swift 5.4 on non-Apple platforms.

kateinoigakukun (approved):

PR #263, opened 2021-10-24
Update dependencies, add support for SwiftWasm 5.5

MaxDesiatov:
I've dropped support for Swift 5.2, and SwiftLint now requires Xcode 12.5.1, which is unavailable on macOS 10.15 with GitHub Actions. This makes it impossible to test with Swift 5.3 on CI, so I'm bumping the required version to Swift 5.4.

~Currently blocked by https://github.com/apple/swift-package-manager/pull/3819~

Resolves https://github.com/swiftwasm/carton/issues/262.


Sources/SwiftToolchain/ToolchainManagement.swift
```diff
@@ -166,6 +167,12 @@ public class ToolchainSystem {
       }.whenComplete($0)
     }
 
+    #if arch(x86_64)
+    let archSuffix = "x86_64"
+    #elseif arch(arm64)
+    let archSuffix = "arm64"
+    #endif
```
kateinoigakukun, line 174: (not big deal)
FYI: `NXGetLocalArchInfo` is more generic way to get the current architecture mode.
You can see how to use it from `arch` command source https://opensource.apple.com/source/cctools/cctools-886/libmacho/arch.c.auto.html

Sources/SwiftToolchain/Manifest.swift
```diff
@@ -122,31 +94,29 @@ struct DumpedManifest: Decodable {
...
-      [DumpedPackageDependencyDescription].self,
+      [PackageDependencyDescription].self,
       forKey: .dependencies
     )
     let products = try container.decode([ProductDescription].self, forKey: .products)
     let targets = try container.decode([TargetDescription].self, forKey: .targets)
     let platforms = try container.decode([PlatformDescription].self, forKey: .platforms)
-    // TODO: Change to non-optional when we move to `Swift 5.4`
-    // `packageKind` is not available in dumped JSON with pre-5.4 compilers.
-    let packageKind = try? container.decode(PackageReference.Kind.self, forKey: .packageKind)
```
kateinoigakukun, line 135: Can carton still parse 5.4 package JSON?

kateinoigakukun (commented):
Almost looks good to me

Sources/SwiftToolchain/Manifest.swift
MaxDesiatov: I haven't tested it, but I'm fine with dropping support for SwiftWasm 5.4 here if that's the easiest way to get 5.5 out.

Sources/SwiftToolchain/ToolchainManagement.swift
MaxDesiatov: This won't work on Linux, right?

kateinoigakukun: Yes, Linux has another mechanism called sysinfo

kateinoigakukun (approved):

Sources/SwiftToolchain/ToolchainManagement.swift
MaxDesiatov: Right, but what would be the benefit of using `sysinfo` on Linux, `NXGetLocalArchInfo` on macOS, and then maybe some third mechanism for Windows (if we ever get there), while `#if arch` seems to achieve the same. Or maybe I'm missing something?

kateinoigakukun: If a user uses a carton built for x64 on arm64 darwin, it downloads x64 toolchain in current implementation. If we get the architecture at runtime, we can avoid such situation and downloads arm64 toolchain even on Rosetta. But I'm not confident which behavior is natural.

MaxDesiatov: I'm not sure if there's much benefit in allow this configuration. I still want to be able to download and test x86_64 toolchains on my M1 hardware. Natural way to do it right now is to just run `carton` under Rosetta. Otherwise we'd have to come up with some CLI flags to make it configurable, which is probably not worth it? I'm ready to change my opinion on this though if a good use case for making this configurable appears.

PR #283, opened 2022-01-01
Use `async`/`await` and actors instead of Combine

MaxDesiatov:
This makes our codebase smaller by ~80 lines and arguably more readable. Also removes OpenCombine dependency.


yonihemi (approved):

yonihemi:
We should remember to add the concurrency warnings to CartonKit.
``` swift
swiftSettings: [
  .unsafeFlags([
    "-Xfrontend", "-warn-concurrency",
    "-Xfrontend", "-enable-actor-data-race-checks",
  ])
]
```
Probably too early now as SwiftPM, TSC Vapor types are not yet annotated with Sendable.

MaxDesiatov:
Not merging this yet, seeing some problems with the watcher not picking up changes...

yonihemi (approved):

yonihemi:
> Not merging this yet, seeing some problems with the watcher not picking up changes...

I've been working with main branch of carton and lately seeing similar issues as well. Not sure if a regression from before but I guess it's not introduced by this PR. I think we can merge and look at the file watcher separately.

PR #288, opened 2022-01-23
Add libsqlite3-dev dependency

### LINKED ISSUE #287: libsqlite3-dev dependency not installed
After running `install_ubuntu_deps.sh` and then attempting to compile Carton by running `swift build -c release`, I encountered this issue:
```
/home/joec/Documents/GitHub/carton/.build/checkouts/swift-tools-support-core/Sources/TSCclibc/include/csqlite3.h:1:10: error: 'sqlite3.h' file not found
#include <sqlite3.h>
         ^
/home/joec/Documents/GitHub/carton/.build/checkouts/swift-tools-support-core/Sources/TSCUtility/SQLite.swift:14:29: error: could not build C module 'CSQLite3'
@_implementationOnly import CSQLite3
```
After installing `libsqlite3-dev`, building carton worked without issue. 
It seems like Carton relies on the `libsqlite3-dev` package, so it should be added to `install_ubuntu_deps.sh`

PR: #288 

Note: I had this issue when running on KDE Neon, which is based on Ubuntu. Since I have not seen this issue before, it seems likely that this issue may be exclusive to KDE Neon, and that package may already be installed on a standard Ubuntu install. That being said, I can't think of any downsides to adding the package to `install_ubuntu_deps.sh`.

SwiftCoderJoe:
Fixes #287 

Not installed by default on all Ubuntu-based distributions.

### DIFF
--- a/install_ubuntu_deps.sh
+++ b/install_ubuntu_deps.sh
@@ -3,9 +3,9 @@
 set -ex
 
 if [ -x "$(command -v sudo)" ]; then
-  sudo apt-get install zlib1g-dev
+  sudo apt-get install zlib1g-dev libsqlite3-dev
 else
-  apt-get install zlib1g-dev
+  apt-get install zlib1g-dev libsqlite3-dev
 fi
 
 BINARYEN_VERSION=105


MaxDesiatov (approved):
Great catch, thanks!

PR #296, opened 2022-02-21
Fix crash for packages with `executableTarget`

yonihemi:
If we run `carton dev` on a package with a Swift 5.4-style `executableTarget`, carton cannot infer the source path and will crash trying to initialize FSWatch with an empty paths array.
```
TSCUtility/FSWatch.swift:46: Precondition failed
```
Fix this issue for executable targets and prevent similar issues with empty arrays.

### DIFF
--- a/Sources/CartonKit/Server/Server.swift
+++ b/Sources/CartonKit/Server/Server.swift
@@ -178,7 +178,7 @@ public actor Server {
         Task { try await self.onChange(changes, configuration) }
       }
       try watcher?.start()
-	}
+  }
   }
 
   private func onChange(_ changes: [AbsolutePath], _ configuration: Configuration) async throws {
--- a/Sources/CartonKit/Server/Server.swift
+++ b/Sources/CartonKit/Server/Server.swift
@@ -178,7 +178,7 @@ public actor Server {
         Task { try await self.onChange(changes, configuration) }
       }
       try watcher?.start()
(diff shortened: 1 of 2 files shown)


Sources/CartonKit/Server/Server.swift
```diff
@@ -172,11 +172,13 @@ public actor Server {
...
-      Task { try await self.onChange(changes, configuration) }
-    }
-    try watcher?.start()
+    if !builder.pathsToWatch.isEmpty {
+      watcher = FSWatch(paths: builder.pathsToWatch, latency: 0.1) { [weak self] changes in
+        guard let self = self, !changes.isEmpty else { return }
+        Task { try await self.onChange(changes, configuration) }
+      }
+      try watcher?.start()
+	}
```
MaxDesiatov, line 181:
```suggestion
  }
```

Sources/CartonKit/Server/Server.swift
```diff
@@ -178,7 +178,7 @@ public actor Server {
         Task { try await self.onChange(changes, configuration) }
       }
       try watcher?.start()
-	}
+  }
```
MaxDesiatov, line 181: sorry, I misaligned the previous suggestion 🤦‍♂️
```suggestion
    }
```

MaxDesiatov (approved):

Issue #277, opened 2021-12-16
Add support for `-Xswiftc` arguments


MaxDesiatov:
Such arguments should be passed to `swift build` invocations directly.

PR #300, opened 2022-03-19
Add -Xswiftc option for each build commands

kateinoigakukun:
Close https://github.com/swiftwasm/carton/issues/277

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,8 +10,9 @@ RUN export DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true && ap
   libncurses5 \
   libsqlite3-0 \
   libsqlite3-dev \
+  libxkbcommon0 \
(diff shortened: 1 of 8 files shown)


yonihemi (dismissed):

MaxDesiatov (dismissed):
Perfect! I was afraid it wouldn't be as easy, I'm glad it turned out to be a small change. Do you think some tests could be added to verify that new options group works as intended?

kateinoigakukun:
BTW, the current style of e2e testing is quite fragile. 
e.g.
- in-place build under fixture, which causes race-condition
  - it also forces each test case to cleanup directories
- AssertExecuteCommand doesn't consider space inside an argument
- AssertExecuteCommand forces how assert the output
- ...

I don't have much time now to address them, but I think we need to restructure testing style.

yonihemi (approved):

Issue #295, opened 2022-02-21
Watcher doesn't see my changes (0.12.2)


gujor:
Hi! I was (am) curious to try this out, but got stuck when the watcher didn't see my file changes.

What I did:
- Followed the install instruction on macOS using brew.
- Used tokamak template.
- Run carton dev, browser started with "Hello, world!".
- Felt this was a great experience so far! Good job!
- Edited and saved the hello message, but it didn't trigger a rebuild or any output from carton / watcher.
- Tried the above on both Big Sur and Monterey, same error.
- Tried previous releases, they worked!
- Started looking through issues, found: https://github.com/swiftwasm/carton/pull/283#issuecomment-1003705593
- Wrote this issue to report that the above mentioned problem still exists :)

yonihemi:
Thanks for reporting.

Jomy10:
I also have this issue.

When I quit the carton dev server using `⌃C`, I get the following output:

```bash
These paths have changed, rebuilding...
- /Users/jonaseveraert/Documents/projects/wasm-game/TokamakApp/Sources/TokamakApp
- /Users/jonaseveraert/Documents/projects/wasm-game/TokamakApp/Sources/TokamakApp
- /Users/jonaseveraert/Documents/projects/wasm-game/TokamakApp/Sources/TokamakApp
Compiling...
/Library/Developer/Toolchains/swift-wasm-5.5.0-RELEASE.xctoolchain/usr/bin/swift build -c debug --product TokamakApp --triple wasm32-unknown-wasi -Xlinker -licuuc -Xlinker -licui18n -Xlinker /Users/jonaseveraert/.carton/static/so_sanitizer.wasm -Xlinker --stack-first
```

Thought this might be useful?

To me it seems like it only rebuild the project when the dev server was quit?

Anyway, great work on making carton, as gujor said:
> Felt this was a great experience so far! Good job!

aargar1:
I am having this problem too. Very annoyning; do you have any hack fixes for the short term?

kateinoigakukun:
It seems this blocking operation violates the Swift Concurrency's assumption of no blocking in async function.
https://github.com/swiftwasm/carton/blob/409c78dd1430d2c2c86bedd6c78275ebb35a208d/Sources/CartonKit/Server/Server.swift#L232

This async task won't be executed due to the violation.
[...]

PR #301, opened 2022-03-25
Add --debug-info flag to carton bundle

kateinoigakukun:


MaxDesiatov (approved):
Thanks!

pedrovgs:
Hey @MaxDesiatov @kateinoigakukun I was about to implement this by myself when I noticed you've just merged a PR with the option we need. Could you please release a new version? the last release published is 12.2.0 Feb 2022 and I'd love to be able to use DWARF info 😃 Thank you so much!

MaxDesiatov:
Thanks! We're planning to tag `carton` 0.13.0 as soon as https://github.com/swiftwasm/carton/issues/295 is fixed, which is a serious regression. It's also going to include support for the future SwiftWasm 5.6 release. I can't promise any dates yet, unfortunately, but I can assure you that tagging a new version is at the top of our list.

You can use your own local build off `main` in the meantime. If that's not possible, let us know here and we'll try to plan accordingly.

PR #302, opened 2022-03-28
Use libSwiftPM 5.6 to parse manifests

### LINKED ISSUE #285: Add support for Swift 5.6 package description format
Starting with Swift 5.6 `swift package dump-package` outputs JSON in format that's incompatible with Swift 5.5 declarations. Ideally, we should be able to support both Swift 5.5 and 5.6 package description formats in `carton`.

yonihemi:
Upgrades libSwiftPM from 5.5 to 5.6 + related dependencies' versions (tools-support-core and swift-argument-parser). Fixes #285.

This changes the parsing strategy to rely on libSwiftPM to directly invoke swiftc and parse a compiled manifests' output. While I expected the output of `swift package dump-package` to be more stable than `PackageDescription`'s JSON output, the opposite is true, and at least between 5.5 and 5.6 that format is backwards compatible. We may have to revisit this if this internal format breaks.

## Previous strategy
 ```mermaid
sequenceDiagram
carton w/ libSwiftPM->>SwiftWasm SwiftPM: (shell) swift package dump-package
SwiftWasm SwiftPM->>SwiftWasm swiftc: build manifest for host platform with host SDK
SwiftWasm swiftc->>SwiftWasm SwiftPM: compiled host manifest
SwiftWasm SwiftPM->>SwiftWasm SwiftPM: parse manifest output
SwiftWasm SwiftPM->>carton w/ libSwiftPM: JSON output
carton w/ libSwiftPM->>carton w/ libSwiftPM: Parse with libSwiftPM
```
## New strategy
 ```mermaid
sequenceDiagram
carton w/ libSwiftPM->>SwiftWasm swiftc: build manifest for host platform with host SDK
SwiftWasm swiftc->>carton w/ libSwiftPM: compiled host manifest
carton w/ libSwiftPM->>carton w/ libSwiftPM: parse manifest output
```

## Alternatives considered:
- Using libSwiftPM 5.6 to parse 5.6 `dump-package` output + copy parts of libSwiftPM 5.5 for parsing older `dump-package` output: This becomes unwieldy fast due to significant code changes with many different types, plus to avoid naming conflicts would need to be isolated to a separate module.
- Go back to our own parser: we'd still need to support at least 2 formats, plus we'd be missing a way to instantiate a proper `Manifest` object.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -60,8 +60,8 @@
         "repositoryURL": "https://github.com/apple/swift-argument-parser.git",
         "state": {
           "branch": null,
-          "revision": "83b23d940471b313427da226196661856f6ba3e0",
-          "version": "0.4.4"
+          "revision": "e394bf350e38cb100b6bc4172834770ede1b7232",
+          "version": "1.0.3"
         }
       },
       {
@@ -78,16 +78,16 @@
         "repositoryURL": "https://github.com/apple/swift-crypto.git",
         "state": {
           "branch": null,
-          "revision": "ddb07e896a2a8af79512543b1c7eb9797f8898a5",
-          "version": "1.1.7"
+          "revision": "3bea268b223651c4ab7b7b9ad62ef9b2d4143eb6",
+          "version": "1.1.6"
         }
       },
       {
         "package": "swift-driver",
         "repositoryURL": "https://github.com/apple/swift-driver.git",
         "state": {
-          "branch": "release/5.5",
-          "revision": "86c54dacd270e0c43374c0cb9b2ceb2924c9ea72",
+          "branch": "release/5.6",
+          "revision": "9982f32f96a2e0e597d1b4a0af4a7e997dc471be",
           "version": null
         }
       },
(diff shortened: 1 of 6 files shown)


MaxDesiatov (approved):
Great, thank you so much! Do I understand correctly that this is compatible with 5.5 toolchains as well?

yonihemi:
> Great, thank you so much! Do I understand correctly that this is compatible with 5.5 toolchains as well?

Correct, tested with both 5.5 and 5.6.

SDGGiesbrecht:
This strategy is compatible with tools version markers as far back as SwiftPM proper supports, but is only compatible with the corresponding _patch_ version of the actual toolchain inside which it expects to reside. If it happens to work between 5.5 and 5.6 by accident, consider yourself lucky, and go ahead and use it for now. Just be aware that you will likely have to switch back to the `dump-package` strategy in the future if you intend to continue supporting multiple toolchains with the same installation of carton. (Having used SwiftPM this way myself since 4.1, I can confirm that it regularly breaks when new toolchains are released.)

PR #304, opened 2022-03-31
Don't strip custom sections when --debug-info

kateinoigakukun:
Since https://github.com/swiftwasm/carton/pull/301, carton started passing `--debuginfo` to wasm-opt, but it still strips all custom sections before it.


MaxDesiatov (approved):
Great catch, thanks!

PR #305, opened 2022-03-31
Bump version to 0.13.0, update `CHANGELOG.md`

MaxDesiatov:
Ideally we'd like to include new JSKit and SwiftWasm releases, but I think it makes more sense to wait for new JSKit release a bit more and to test things more thoroughly. Also, the watcher bug is still not fixed.

On the other hand, `--debug-info` has been highly requested by our users, so nothing prevents us from releasing 0.13.0 right now, and 0.14.0 with new JSKit/Tokamak and SwiftWasm 5.6 later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ SwiftPM 5.6 API in preparation for the imminent SwiftWasm 5.6 release.
 
 This version of `carton` now ships with JavaScriptKit 0.13.0 runtime.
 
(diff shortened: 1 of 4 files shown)


MaxDesiatov:
Moving it to draft for now, thinking maybe we could tag SwiftWasm 5.6.0 and new JSKit after all?

MaxDesiatov:
Currently waiting for https://github.com/swiftwasm/JavaScriptKit/pull/177

kateinoigakukun (approved):

PR #307, opened 2022-04-01
Fix watcher blocked by Vapor `run()`

kateinoigakukun:
Actor isolated methods must not block the current thread, so Vapor
server's start method is not compatible in Concurrency context.

Resolve https://github.com/swiftwasm/carton/issues/295

### DIFF
--- a/Sources/CartonKit/Server/Server.swift
+++ b/Sources/CartonKit/Server/Server.swift
@@ -227,9 +227,26 @@ public actor Server {
   }
 
   /// Blocking function that starts the HTTP server.
-  public func run() throws {
-    defer { app.shutdown() }
-    try app.run()
+  public nonisolated func run() async throws {
+    // Explicitly hop to another thread to avoid blocking the thread that is running the actor executor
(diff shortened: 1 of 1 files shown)


Sources/CartonKit/Server/Server.swift
```diff
@@ -227,9 +227,26 @@ public actor Server {
   }
 
   /// Blocking function that starts the HTTP server.
-  public func run() throws {
-    defer { app.shutdown() }
-    try app.run()
+  public nonisolated func run() async throws {
+    // Explicitly hop to another thread to avoid blocking the thread that is running the actor executor
+    let _: () = try await withCheckedThrowingContinuation { continuation in
```
j-f1, line 232: Can this be…
```suggestion
    try await withCheckedThrowingContinuation { continuation in
```

…since the `Void` value is not used?

MaxDesiatov (commented):
Nice catch, but I'm also curious if `let _: () = ` is required here?

aargar1 (dismissed):
Great

MaxDesiatov (approved):

Issue #313, opened 2022-04-11
“carton test” fails to run as of 0.14.0


SDGGiesbrecht:
```
$ carton test
Directory doesn't exist or contains outdated polyfills: /root/.carton/static
Polyfills archive successfully downloaded from https://github.com/swiftwasm/carton/releases/download/0.13.0/static.zip
Unpacking the archive: /root/.carton/static.zip
Entrypoint integrity verified: /root/.carton/static/test.js
- checking Swift compiler path: /root/.carton/sdk/wasm-5.6.0-RELEASE/usr/bin/swift
Inferring basic settings...
- swift executable: /root/.carton/sdk/wasm-5.6.0-RELEASE/usr/bin/swift
SwiftWasm Swift version 5.6 (swiftlang-5.6.0)
Target: x86_64-unknown-linux-gnu

Parsing package manifest:
[...]

Building the test bundle before running the test suite...
Compiling...
/root/.carton/sdk/wasm-5.6.0-RELEASE/usr/bin/swift build -c debug --product [...]Tests -Xswiftc -color-diagnostics --triple wasm32-unknown-wasi -Xswiftc -Xclang-linker -Xswiftc -mexec-model=reactor -Xlinker --export=main
[...]
Build complete! [...]

`swift` process finished successfully
`swift build` completed [...]

Running the test bundle with wasmer:
Running...
wasmer [...]/.build/wasm32-unknown-wasi/debug/[...]Tests.wasm

Process failed and produced following output: 
Error: Process failed with non-zero exit status and following output:
 and following error output:
error: failed to run `[...]/.build/wasm32-unknown-wasi/debug/[...]Tests.wasm`
╰─▶ 1: No export `_start` found in the module.
       Similar functions found: `main`, `_initialize`.
       Try with: wasmer [...]/.build/wasm32-unknown-wasi/debug/[...]Tests.wasm -i main
```

The two mismatches I notice in the log are:
- `-Xlinker --export=main` vs `No export  ̀_start ̀ found in the module.`
- polyfills are reported as coming from 0.13.0, despite the docker image being 0.14.0 and the default Swift resolving to 5.6.

Is there a way tell Carton to pass `-i main` as a workaround?

MaxDesiatov:
> polyfills are reported as coming from 0.13.0, despite the docker image being 0.14.0 and the default Swift resolving to 5.6.

[...]

PR #310, opened 2022-04-06
Several fixes for 5.6 toolchain

kateinoigakukun:
- Reactor model support
- Update sanitizer support object

This is required to resolve https://github.com/TokamakUI/Tokamak/pull/475

PR #314, opened 2022-04-11
Fix "No export `_start` found in the module" error in `carton test` 

kateinoigakukun:
This enables reactor model only when not building as a command, fixes https://github.com/swiftwasm/carton/issues/313
The breakage happened and was silenced in https://github.com/swiftwasm/carton/pull/310


MaxDesiatov (approved):
Thank you! 🙏

PR #323, opened 2022-04-27
Override clock_res_get function from wasmer/wasi-js to fix memory issue

### LINKED ISSUE #321: Apply `clock_res_get` patch in all entrypoints
This is an issue with `@wasmer/wasi` that we should work around on our side. See a comment at https://github.com/swiftwasm/WasmTransformer/blob/a6d18ccfe15bb0cc93799296909101236e47ad3b/Fixtures/index.js#L21 for more details.

fjtrujy:
This PRs just basically apply the same patch that was done in the WasmTransformer repo. https://github.com/swiftwasm/WasmTransformer/blob/main/Fixtures/index.js#L4-L27

It is overriding the implementation of `clock_res_get` in order to solve a memory issue.
It is applied to all the entry points that `carton` has.

This PR should closes #321 .

Thanks

### DIFF
--- a/entrypoint/bundle.js
+++ b/entrypoint/bundle.js
@@ -46,6 +46,24 @@ const wasi = new WASI({
   },
 });
 
+const wrapWASI = (wasiObject) => {
+  // PATCH: @wasmer-js/wasi@0.x forgets to call `refreshMemory` in `clock_res_get`,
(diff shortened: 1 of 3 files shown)


kateinoigakukun (approved):

MaxDesiatov (approved):
Thanks!

PR #328, opened 2022-05-10
Add support for running tests using NodeJS

AntonioCandinho:
### :tophat: What is the goal?

Implement NodeJS as another test runner. 

### :page_facing_up: How is it being implemented?

In this PR, I include the following changes:

 - Create an abstraction for the different Test Runners (Wasmer, browser and now Node)
 - Implement the new Node test runner (quite similar to the Wasmer one)
 - Unify Javascript client code, so we can make sure we apply the same patches and load in the same way all the different targets
 - Create a new entry point for testing with node
 - Add that new entry point to the static bundle
 
### :eyes: Any consideration?

The Node test runner does not use the TestsParser as it heavily impacts execution time. I will try to figure out what's going on later.
 
### :white_check_mark: How can it be tested?

Testing is automated 🤖 . You can also check this in your own project by running:

```bash
carton test --environment node
```


Sources/CartonCLI/Commands/Test.swift
```diff
@@ -48,7 +50,8 @@ struct Test: AsyncParsableCommand {
   @Argument(help: "The list of test cases to run in the test suite.")
   var testCases = [String]()
 
-  @Option(help: "Environment used to run the tests, either a browser, or command-line Wasm host.")
+  @Option(
+    help: "Environment used to run the tests, either a browser, node, or command-line Wasm host.")
```
MaxDesiatov, line 54: I hope string interpolation works in this context. That would help us generate the list of available environments automatically.

```suggestion
  @Option(
    help: "Environment used to run the tests. Available values: \(Environment.allCases.joined(separator: ","))")
```

Sources/CartonCLI/Commands/Test.swift
```diff
@@ -69,49 +72,44 @@ struct Test: AsyncParsableCommand {
...
   }
+
+  private func buildFlavor() -> BuildFlavor {
+    BuildFlavor(
+      isRelease: release,
+      environment: environment.destination,
+      sanitize: sanitize,
+      swiftCompilerFlags: buildOptions.swiftCompilerFlags
+    )
+  }
```
MaxDesiatov, line 114: coding style nit: I realize this was a function before, but let's make this a property, not a function, while we're at it.
```suggestion
  private var buildFlavor: BuildFlavor {
    BuildFlavor(
      isRelease: release,
      environment: environment.destination,
      sanitize: sanitize,
      swiftCompilerFlags: buildOptions.swiftCompilerFlags
    )
  }
```

MaxDesiatov (commented):
Overall this looks great to me, just a couple of nits

entrypoint/common.js
```diff
@@ -0,0 +1,139 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { SwiftRuntime } from "javascript-kit-swift";
+import { WASI } from "@wasmer/wasi";
+import { WasmFs } from "@wasmer/wasmfs";
```
kateinoigakukun, line 17: We don't need to use Wasmer's polyfill on Node.js because it also has builtin WASI implementation.

kateinoigakukun (changes requested):

Dockerfile
```diff
@@ -24,6 +24,12 @@ RUN mkdir -p $CARTON_ROOT/sdk && \
 
 COPY . carton/
 
+ENV NODE_VERSION=18.1.0 
+
+RUN curl -fsSLO --compressed "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-x64.tar.xz" && \
+    tar -xJf "node-v$NODE_VERSION-linux-x64.tar.xz" -C /usr/local --strip-components=1 --no-same-owner && \
+    ln -s /usr/local/bin/node /usr/local/bin/nodejs
```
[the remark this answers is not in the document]
AntonioCandinho: Nope, getting rid of it... 😄

entrypoint/common.js
AntonioCandinho: I am afraid that the last time I tried running with the native Node WASI, the tests crashed. I think it could maybe be something interesting in the future, though.  My concerns doing that are:

- Node's WASI is still in an experimental stage
- Tests will use a different WASI implementation than the one used in production. Using the Wasmer's polyfill in Node has allowed us in the past to detect errors like the `clock_rest_get` memory refresh (btw kudos to you, @kateinoigakukun, for the help there)

What do you think?

AntonioCandinho:
Thank you so much for your feedback! I've already applied most of the suggestions 😄

entrypoint/common.js
kateinoigakukun: > - Node's WASI is still in an experimental stage

My impression is Node's WASI is more stable, maintained, and well-tested than wasmer's one.
For example, wasmer's one doesn't support reactor model, and it's recently replaced with an entirely new implementation since version 1.x but it's still not production-ready quality. So I prefer Node.js's one personally.

> Tests will use a different WASI implementation than the one used in production.

OK, it seems reasonable to keep wasmer's WASI implementation while using it in the browser environment.

Let's investigate the crash and add options to switch WASI backends in another PR.

MaxDesiatov: @kateinoigakukun do you have anything that you'd like to see addressed in this specific PR? If not, can it get your approval then?

kateinoigakukun (dismissed):

Sources/CartonCLI/Commands/Test.swift
```diff
@@ -48,7 +53,8 @@ struct Test: AsyncParsableCommand {
   @Argument(help: "The list of test cases to run in the test suite.")
   var testCases = [String]()
 
-  @Option(help: "Environment used to run the tests, either a browser, or command-line Wasm host.")
+  @Option(
+    help: "Environment used to run the tests: \(Environment.allCasesNames.joined(separator: ", "))")
```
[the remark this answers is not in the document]
MaxDesiatov: @AntonioCandinho would you mind pushing this small update? I don't think I have a permission to do that myself. Otherwise it looks ready for merging to me.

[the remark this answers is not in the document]
AntonioCandinho: Done! Thnks 🙏

MaxDesiatov (approved):

PR #330, opened 2022-05-10
Serve all available resources with `dev`/`bundle`

MaxDesiatov:
Resolves https://github.com/swiftwasm/carton/issues/175.

I initially thought we should parse `Package.swift` manifests of the whole dependency tree to collect paths to resources from them, but now I'm not even sure that SwiftPM provides an API for this.

Much simpler solution is to serve with `dev` and copy with `bundle` all directories with `.resources` suffixes in the build directory. I think it's quite impossible to stumble upon unrelated directories with this approach, while it still resolves the issue as intended, in my opinion.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -8,7 +8,7 @@ disabled_rules:
   - opening_brace
   - todo
 
-line_length: 100
+line_length: 120
 
 function_body_length: 50
 
--- a/Sources/CartonCLI/Commands/Bundle.swift
+++ b/Sources/CartonCLI/Commands/Bundle.swift
@@ -155,10 +155,10 @@ struct Bundle: AsyncParsableCommand {
     )
 
     let manifest = try toolchain.manifest.get()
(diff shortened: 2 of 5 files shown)


.swiftlint.yml
```diff
@@ -8,7 +8,7 @@ disabled_rules:
   - opening_brace
   - todo
 
-line_length: 100
+line_length: 120
```
MaxDesiatov, line 11: I find that the limit of 100 is too narrow in practice, there are some comments with links that don't fit.

MaxDesiatov, line 11: I find that the limit of 100 is too narrow in practice, there are some comments with links that don't fit.

yonihemi (approved):

yonihemi:
> I initially thought we should parse `Package.swift` manifests of the whole dependency tree to collect paths to resources from them, but now I'm not even sure that SwiftPM provides an API for this.

It is possible, but we already saw libSwiftPM does not have a stable public API. I think your solution is much better for now.

PR #334, opened 2022-05-11
Embed static.zip in Swift code instead of downloading from remote

kateinoigakukun:
This is short-term mitigation of release-process until https://github.com/swiftwasm/JavaScriptKit/issues/191


j-f1:
I’m also curious — what’s the most efficient way to store data in a Swift binary? Base64 data of course takes up about ⅓ more space than the equivalent raw data. In particular, are the arrays currently used for the hashes serialized to raw bytes in the binary, or would it perhaps make sense to store the encoded bytes in a C module?

kateinoigakukun:
@j-f1 Yes, Base64 is not efficient way... I initially tried to store it as byte array literal, but too many literal elements will crash the type-checker.

j-f1 (approved):

PR #342, opened 2022-05-19
Re-read custom `index.html` on updates

MaxDesiatov:
We were reading custom `index.html` only once on launch, which meant people had to restart `carton dev` every time they've updated `index.html` during the build process to see changes they've made.

Resolves https://github.com/swiftwasm/carton/issues/341.

cc @ephemer


Package.swift
```diff
@@ -5,7 +5,7 @@ import PackageDescription
 
 let package = Package(
   name: "carton",
-  platforms: [.macOS("10.15.4")],
+  platforms: [.macOS(.v12)],
```
MaxDesiatov, line 8: Had to bump this because Vapor puts a macOS 12 requirement for `async` routes. Can rewrite those new routes with `EventLoopFuture` instead, but I've been meaning to bump the package manifest to Swift 5.6, which requires macOS 12 to build anyway.

@ephemer please let me know if support for earlier versions of macOS is critical for you.

Sources/CartonKit/Server/HTML.swift
```diff
@@ -25,13 +25,19 @@ public struct HTML {
...
+    return try await request.eventLoop.makeSucceededFuture(.init(
       status: .ok, headers: headers, body: .init(string: value)
-    ))
+    )).get()
+  }
+
+  public static func customIndexPath(from path: String, on fileSystem: FileSystem) -> AbsolutePath {
+    path.isAbsolutePath ?
+      AbsolutePath(path) :
+      AbsolutePath(localFileSystem.currentWorkingDirectory!, path)
```
kateinoigakukun, line 40:
```swift
AbsolutePath(path, relativeTo: localFileSystem.currentWorkingDirectory!)
```

is equivalent to this. And also, I think this method doesn't need to be extracted as a method in here since it's not specific to HTML thing.

Package.swift
kateinoigakukun: I think the async things are not an essential part of this PR, and the [`@available` declaration in vapor](https://github.com/vapor/vapor/blob/deef482f6f309db01666ac99c8b22b6478ef2be5/Sources/Vapor/Concurrency/RoutesBuilder%2BConcurrency.swift#L4) can be relaxed to 10.15.
I think bumping up the minimum supported OS version is critical for production users, so we should discuss it in another PR.

kateinoigakukun (approved):

PR #347, opened 2022-05-20
Fix watcher breakage on failed builds

### LINKED ISSUE #339: `carton dev` stops rebuilding/reloading after error

MaxDesiatov:
After switching to `async`/`await` the value of `isBuildCurrentlyRunning` wasn't correctly reset, which led to issues. By resetting the variable in `defer` we guarantee that it will be reset even when errors are thrown.

Resolves #339.

### DIFF
--- a/Tests/Fixtures/NodeJSKitTest/.gitignore
+++ b/Tests/Fixtures/NodeJSKitTest/.gitignore
@@ -7,3 +7,4 @@ DerivedData/
 .swiftpm/config/registries.json
 .swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
 .netrc
+.vscode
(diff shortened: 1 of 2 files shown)


Tests/Fixtures/NodeJSKitTest/.gitignore
```diff
@@ -7,3 +7,4 @@ DerivedData/
 .swiftpm/config/registries.json
 .swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
 .netrc
+.vscode
```
MaxDesiatov, line 10: Just a small dev experience change to avoid including `launch.json` included, didn't know where else to squeeze this small change in.

kateinoigakukun: `~/.config/git/ignore` or `.git/info/exclude`? but it's ok to include it here.

kateinoigakukun (approved):

PR #352, opened 2022-05-23
Add `<!DOCTYPE html>` to default `index.html`

MaxDesiatov:
As suggested by @carson-katri the prevents browsers from using quirks mode by default.


carson-katri:
I just opened a PR for this too #351 😂  You can close whichever one you want.

PR #355, opened 2022-05-27
Clean up integration tests and fixtures

kateinoigakukun:
- Stop using Tokamak app fixture for every test case to avoid long build
- Stop writing file in Fixture directory during testing


.swiftlint.yml
```diff
@@ -7,10 +7,10 @@ disabled_rules:
...
   - opening_brace
   - todo
-
-line_length: 120
-
-function_body_length: 50
+  - function_parameter_count
+  - line_length
+  - function_body_length
+  - file_length
```
MaxDesiatov, line 13: Any specific need for this change? I think we had pretty reasonable limits here and it doesn't look like this PR is exceeding them.

MaxDesiatov (commented):
Overall this is a good and overdue refactoring, but do I understand correctly that no tests run in parallel, so the run time of the test suite is roughly the same?

kateinoigakukun:
Yes, parallel execution makes tests slower in CI due to limited resources, so I disabled it in CI.

.swiftlint.yml
kateinoigakukun: SwiftLint action somehow reports errors https://github.com/swiftwasm/carton/actions/runs/2397807140 for the whole code base not only for the diff.

MaxDesiatov: Ah, that's fair. Is it a required check? If so, let's make SwiftLint not required and maybe revert these config changes and we'll cleanup warnings in a separate PR?

kateinoigakukun: ok

MaxDesiatov (approved):
Thanks! 🙏

PR #356, opened 2022-05-28
Fix linter error in `Builder.swift`

MaxDesiatov:


Sources/SwiftToolchain/Builder.swift
```diff
@@ -80,23 +80,23 @@ public final class Builder {
...
-    let binary = try self.fileSystem.readFileContents(self.mainWasmPath)
+    let binary = try fileSystem.readFileContents(mainWasmPath)
 
     let transformStarted = Date()
     var inputBinary = binary.contents
     for transformer in transformers {
       var input = InputByteStream(bytes: inputBinary)
       var writer = InMemoryOutputWriter(reservingCapacity: inputBinary.count)
-      try! transformer(&input, &writer)
+      try transformer(&input, &writer)
```
MaxDesiatov, line 90: This is the line that was causing SwiftLint check to fail. The rest of the changes are applied by the formatter.

kateinoigakukun (approved):

PR #358, opened 2022-05-30
Add --bundle-path option to `carton test`

kateinoigakukun:
This allows users to build their projects in their responsibilities with custom options.

### DIFF
--- a/Sources/CartonCLI/Commands/Test.swift
+++ b/Sources/CartonCLI/Commands/Test.swift
@@ -56,9 +56,8 @@ struct Test: AsyncParsableCommand {
   )
   var host = "127.0.0.1"
 
-  @Flag(name: .customLong("skip-build"),
-        help: "Skip building the test target")
-  var shouldSkipBuilding: Bool = false
+  @Option(help: "Use the given bundle instead of building the test target")
+  var bundlePath: String?
 
   @OptionGroup()
   var buildOptions: BuildOptions
@@ -76,8 +75,8 @@ struct Test: AsyncParsableCommand {
     let terminal = InteractiveWriter.stdout
     let toolchain = try await Toolchain(localFileSystem, terminal)
     let bundlePath: AbsolutePath
-    if shouldSkipBuilding {
-      bundlePath = try toolchain.getTestProduct(flavor: buildFlavor).artifactPath
+    if let preBundlePath = self.bundlePath {
+      bundlePath = AbsolutePath(preBundlePath, relativeTo: localFileSystem.currentWorkingDirectory!)
       guard localFileSystem.exists(bundlePath) else {
         terminal.write(
           "No built binary found in \(bundlePath)\n",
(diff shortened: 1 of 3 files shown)


yonihemi (approved):

Sources/CartonCLI/Commands/Test.swift
```diff
@@ -56,6 +56,9 @@ struct Test: AsyncParsableCommand {
   )
   var host = "127.0.0.1"
 
+  @Option(help: "Use the given bundle instead of building the test target")
+  var bundlePath: String?
```
MaxDesiatov, line 60: Would you consider something like `prebuiltBinaryPath` here? It's not clear what exactly "bundle" means in this context. It took me some time to remember that XCTest names these as test bundles, but I think "prebuilt binary" or something like that is much more clear.

kateinoigakukun: Yeah, actually the produced wasm binary is not "bundle" in darwin terms, and I agree it's unclear here...

`--prebuilt-binary-path` sounds better to me but not the best since "binary" is a little bit ambiguous.
But let's go ahead with `--prebuilt-binary-path` and rename it after we will come up with a good idea.

MaxDesiatov: `--prebuilt-test-bundle-path` could also work maybe? But still confusing in the context of `carton bundle`, so maybe `--prebuilt-binary-path` really is the best currently known option?

MaxDesiatov: Another way could be to pass it to `carton test directly`, without options: `carton test test.wasm`

kateinoigakukun: A little bit long, but sounds better 👍

MaxDesiatov: WDYT about passing to `carton test` directly, like `carton test test.wasm`?

kateinoigakukun: I thought it first, but positional arguments are already used as `testCases`

Sources/CartonCLI/Commands/Test.swift
```diff
@@ -71,7 +74,19 @@ struct Test: AsyncParsableCommand {
   func run() async throws {
     let terminal = InteractiveWriter.stdout
     let toolchain = try await Toolchain(localFileSystem, terminal)
-    let bundlePath = try await toolchain.buildTestBundle(flavor: buildFlavor)
+    let bundlePath: AbsolutePath
+    if let preBundlePath = self.prebuiltTestBundlePath {
+      bundlePath = AbsolutePath(preBundlePath, relativeTo: localFileSystem.currentWorkingDirectory!)
+      guard localFileSystem.exists(bundlePath) else {
+        terminal.write(
+          "No prebuilt binary found in \(bundlePath)\n",
```
MaxDesiatov, line 82: nit
```suggestion
          "No prebuilt binary found at \(bundlePath)\n",
```

MaxDesiatov:
@kateinoigakukun sorry I'm too late with this suggestion, but can you also update `README.md` section about `carton test` with a sentence or two about this new option? Thanks!

PR #359, opened 2022-05-30
Add option to disable wasm optimizations

ephemer:
I am trying to implement an `esbuild` plugin that calls into `carton`. My plan was to call `carton bundle --debug` to get a quick development build and then extract that command's output for use in our esbuild project.

Unfortunately, `carton bundle` still runs `wasm-opt` even on a debug build, which takes 5-10s for our project. The only alternative I can see is `carton dev`, but that runs the dev server / watcher, which we don't want either for a one-off build.

Since changing the behaviour of `carton bundle --debug` to _not_ run `wasm-opt` may cause issues with backwards compatibility, @MaxDesiatov suggested we add a command line option `--wasm-optimizations {size, none}` instead, allowing users to specifically opt out of this behaviour.

The current PR implements the new `--wasm-optimizations` option, as can be seen in the screenshot below:

image

### DIFF
--- a/Sources/CartonCLI/Commands/Bundle.swift
+++ b/Sources/CartonCLI/Commands/Bundle.swift
@@ -48,7 +48,12 @@ struct Bundle: AsyncParsableCommand {
 
   @Option(
     name: .long,
-    help: "Which optimizations to apply to the .wasm binary output.\n Available values: \(WasmOptimizations.allCases.map { $0.rawValue }.joined(separator: ", "))"
+    help: """
+        Which optimizations to apply to the .wasm binary output.
+        Available values: \(
+            WasmOptimizations.allCases.map { $0.rawValue }.joined(separator: ", ")
+        )
+        """
   )
   var wasmOptimizations: WasmOptimizations = .size
 
--- a/Sources/CartonCLI/Commands/Bundle.swift
+++ b/Sources/CartonCLI/Commands/Bundle.swift
@@ -49,11 +49,11 @@ struct Bundle: AsyncParsableCommand {
   @Option(
     name: .long,
     help: """
-        Which optimizations to apply to the .wasm binary output.
-        Available values: \(
-            WasmOptimizations.allCases.map { $0.rawValue }.joined(separator: ", ")
-        )
-        """
+    Which optimizations to apply to the .wasm binary output.
+    Available values: \(
+      WasmOptimizations.allCases.map(\.rawValue).joined(separator: ", ")
+    )
+    """
   )
   var wasmOptimizations: WasmOptimizations = .size
 
@@ -101,7 +101,7 @@ struct Bundle: AsyncParsableCommand {
     let bundleDirectory = AbsolutePath(localFileSystem.currentWorkingDirectory!, "Bundle")
     try localFileSystem.removeFileTree(bundleDirectory)
     try localFileSystem.createDirectory(bundleDirectory)
-    
+
     let wasmOutputFilePath = AbsolutePath(bundleDirectory, "main.wasm")
 
     if wasmOptimizations == .size {
--- a/Sources/CartonCLI/Commands/Bundle.swift
+++ b/Sources/CartonCLI/Commands/Bundle.swift
@@ -26,6 +26,10 @@ private let dependency = Entrypoint(
   sha256: bundleEntrypointSHA256
 )
 
+enum WasmOptimizations: String, CaseIterable, ExpressibleByArgument {
+  case size, none
+}
+
 struct Bundle: AsyncParsableCommand {
   @Option(help: "Specify name of an executable product to produce the bundle for.")
   var product: String?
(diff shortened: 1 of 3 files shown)


Sources/CartonCLI/Commands/Bundle.swift
```diff
@@ -42,6 +46,10 @@ struct Bundle: AsyncParsableCommand {
   @Flag(help: "Emit names and DWARF sections in the .wasm file.")
   var debugInfo: Bool = false
 
+  @Option(name: .long,
+          help: "Which optimizations to apply to the .wasm binary output.\nAvailable values: \(WasmOptimizations.allCases.map { $0.rawValue }.joined(separator: ", "))")
```
MaxDesiatov, line 50:
```suggestion
  @Option(
    name: .long,
    help: "Which optimizations to apply to the .wasm binary output.\nAvailable values: \(
      WasmOptimizations.allCases.map { $0.rawValue }.joined(separator: ", ")
    )"
  )
```

MaxDesiatov (changes requested):
Would you mind adding a test for both option values in `BundleCommandTests.swift`? https://github.com/swiftwasm/carton/blob/70b06afad8d46bcf4a9e0e28cf3978bbeef94f0f/Tests/CartonCommandTests/BundleCommandTests.swift#L54

MaxDesiatov:
One more request here: please update `README.md` section that describes commands and add a description of this new option. Thanks!

Sources/CartonCLI/Commands/Bundle.swift
ephemer: Hi @MaxDesiatov the code does not build like this. I don't know why but it appears putting `WasmOptimizations.allCases.map { $0.rawValue }.joined(separator: ", ")` onto a new line like this breaks something to do with Swift's parser. I believe it's a Swift bug

ephemer: Actually, it seems that any newline characters within `"`-delimited strings will break the parsing. I'm not sure this is even a bug but may be by design. So I need to revert this change otherwise the code will not compile.

MaxDesiatov: I think you'll just have to use `"""`-delimited multiline strings for multiline interpolation to work.

Sources/CartonCLI/Commands/Bundle.swift
```diff
@@ -42,6 +46,12 @@ struct Bundle: AsyncParsableCommand {
   @Flag(help: "Emit names and DWARF sections in the .wasm file.")
   var debugInfo: Bool = false
 
+  @Option(
+    name: .long,
+    help: "Which optimizations to apply to the .wasm binary output.\n Available values: \(WasmOptimizations.allCases.map { $0.rawValue }.joined(separator: ", "))"
```
MaxDesiatov, line 51: Still wondering if multiline string would allow us to use multiline interpolation here?

```suggestion
    help: """
      Which optimizations to apply to the .wasm binary output.\n Available values: \(
        WasmOptimizations.allCases.map { $0.rawValue }.joined(separator: ", ")
      )
    """
```

ephemer:
Hi @MaxDesiatov, I have added a test and updated the README. If you can point me to how you were made aware of the linting error you saw in the first commit I would be happy to check and fix that too

Sources/CartonCLI/Commands/Bundle.swift
ephemer: The latest commit builds, is that what you had in mind?

MaxDesiatov (commented):
Great, thanks! I'm only waiting for CI to pass and maybe for another review if anyone from the team has a moment to have a glance at this.

MaxDesiatov (approved):

ephemer:
After discussing this with @MaxDesiatov a bit further on Discord, I realised that it'd also be useful for us to disable the Safari transformations as well (in order to speed up the build further). With that in mind, I would like to make an alternative suggestion to the approach taken in this PR:

**Instead of adding a flag `--wasm-optimizations {size,none}`, I would like to suggest `--disable-transformations [size,safari,debug-info,all]` instead.** The `size` option would make the current PR redundant, and `debug-info` would also make the `--debug-info` flag redundant (but we can keep it for now for backwards compat of course).

Since it does make sense to run the transformations by default in most cases (AFAIK in all cases with `carton bundle`), a useful API would allow users to selectively (or entirely, via `all`) disable various transformations that might run during the bundling step. As more and more transformations are added over time, it'd be useful to disable one or more of them via this command line option.

Since all transformations currently run by default, adding a `--enable-transformations` option would be redundant at this time, but it could be a useful possibility in the future (thanks @MaxDesiatov for this suggestion).

I understand @kateinoigakukun has done significant work on the transformations pipeline, so I'd like to get your feedback here as well. Since I need the option for my own work, I am happy to create a draft PR outlining the API of suggestion.

ephemer:
@kateinoigakukun the only minor issue I see with the alternative name for the `safari` transformation is that it is referred to exclusively by `carton` in the user-facing messaging as e.g. "Binary transformation for Safari compatibility completed in 1.26 seconds".

So I think for discoverability it would be clearer if we changed that message to "Int64-lowering transformation (for Safari compatibility) completed in 1.26 seconds". Would that work for you and @MaxDesiatov?

MaxDesiatov:
Thank you for noticing this detail, your proposed solution LGTM!

MaxDesiatov:
Ok, let's get an approval from Yuta then before this is merged.

kateinoigakukun (approved):

MaxDesiatov (approved):

MaxDesiatov:
Thanks!

PR #362, opened 2022-06-09
Headless test runner

kateinoigakukun:
## Motivation 

`carton test --environment defaultBrowser` only works on macOS or X Window systems, so tests using browser APIs cannot be run on CI easily. Add a headless test runner based on WebDriver.

## Changed

- Add `WebDriverClient` module, which is a thin client of W3C WebDriver: https://w3c.github.io/webdriver/
- Add `--headless` option to `carton test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ either: `wasmer`, `node` or `defaultBrowser`. Code that depends on
 the test run will not succeed, since JavaScript environment is not available with `--environment wasmer`.
 If you want to run your test suite on CI or without GUI but on browser, you can pass `--headless` flag.
 It enables [WebDriver](https://w3c.github.io/webdriver/)-based headless browser testing. Note that you
-need to install a WebDriver in `PATH` before running tests.
+need to install a WebDriver executable in `PATH` before running tests.
 You can use the command with a prebuilt test bundle binary instead of building it in carton by passing
 `--prebuilt-test-bundle-path <your binary path>`.
 
--- a/README.md
+++ b/README.md
@@ -93,6 +93,9 @@ either: `wasmer`, `node` or `defaultBrowser`. Code that depends on
 [JavaScriptKit](https://github.com/swiftwasm/JavaScriptKit) should pass either `--environment node` or
 `--environment defaultBrowser` options, depending on whether it needs Web APIs to work. Otherwise
 the test run will not succeed, since JavaScript environment is not available with `--environment wasmer`.
+If you want to run your test suite on CI or without GUI but on browser, you can pass `--headless` flag.
+It enables [WebDriver](https://w3c.github.io/webdriver/)-based headless browser testing. Note that you
+need to install a WebDriver in `PATH` before running tests.
 You can use the command with a prebuilt test bundle binary instead of building it in carton by passing
 `--prebuilt-test-bundle-path <your binary path>`.
 
--- a/Sources/CartonCLI/Commands/Test.swift
+++ b/Sources/CartonCLI/Commands/Test.swift
@@ -45,6 +45,9 @@ struct Test: AsyncParsableCommand {
   )
   private var environment = Environment.wasmer
 
+  // It implemented as a separate flag instead of a `--environment` variant because `--environment`
+  // is designed to accept specific browser names in the future like `--environment firefox`. 
+  // Then `--headless` should be able to be used with `defaultBrowser` and other browser values.
   @Flag(help: "When running browser tests, run the browser in headless mode")
   var headless: Bool = false
 
--- a/Sources/CartonCLI/Commands/Test.swift
+++ b/Sources/CartonCLI/Commands/Test.swift
@@ -22,6 +22,10 @@ extension Environment: ExpressibleByArgument {}
 
 extension SanitizeVariant: ExpressibleByArgument {}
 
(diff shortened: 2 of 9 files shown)


MaxDesiatov:
Is this new option compatible with other `--environment` values? If not, maybe it should be implemented as `--environment webdriver`?

kateinoigakukun:
@MaxDesiatov I made it as a new flag instead of a `--environment` variant because I thought `--environment` is designed to accept specific browser names in the future like `--environment firefox` and then `--headless` should be able to be used with `defaultBrowser` and other browser values.

MaxDesiatov:
That makes sense, thanks!

README.md
```diff
@@ -93,6 +93,9 @@ either: `wasmer`, `node` or `defaultBrowser`. Code that depends on
 [JavaScriptKit](https://github.com/swiftwasm/JavaScriptKit) should pass either `--environment node` or
 `--environment defaultBrowser` options, depending on whether it needs Web APIs to work. Otherwise
 the test run will not succeed, since JavaScript environment is not available with `--environment wasmer`.
+If you want to run your test suite on CI or without GUI but on browser, you can pass `--headless` flag.
+It enables [WebDriver](https://w3c.github.io/webdriver/)-based headless browser testing. Note that you
+need to install a WebDriver in `PATH` before running tests.
```
MaxDesiatov, line 98:
```suggestion
need to install a WebDriver executable in `PATH` before running tests.
```

Sources/CartonCLI/Commands/Test.swift
```diff
@@ -41,6 +45,12 @@ struct Test: AsyncParsableCommand {
   )
   private var environment = Environment.wasmer
 
+  // It implemented as a separate flag instead of a `--environment` variant because `--environment`
+  // is designed to accept specific browser names in the future like `--environment firefox`. 
+  // Then `--headless` should be able to be used with `defaultBrowser` and other browser values.
```
MaxDesiatov, line 50: Sorry, I wanted to suggest a doc comment, but used a regular comment delimiter for the previous suggestion.
```suggestion
  /// It is implemented as a separate flag instead of a `--environment` variant because `--environment`
  /// is designed to accept specific browser names in the future like `--environment firefox`. 
  /// Then `--headless` should be able to be used with `defaultBrowser` and other browser values.
```

Sources/CartonCLI/Commands/TestRunners/BrowserTestRunner.swift
```diff
@@ -12,39 +12,165 @@
...
 import TSCBasic
+import WebDriverClient
 
 private enum Constants {
   static let entrypoint = Entrypoint(fileName: "test.js", sha256: testEntrypointSHA256)
 }
 
+enum BrowserTestRunnerError: Error, CustomStringConvertible {
+  case invalidRemoteURL(String)
+  case failedToFindDriver
```
MaxDesiatov, line 31: nit: let's make naming more explicit here
```suggestion
  case failedToFindWebDriver
```

Sources/CartonCLI/Commands/TestRunners/BrowserTestRunner.swift
```diff
@@ -12,39 +12,165 @@
...
+
+  var description: String {
+    switch self {
+    case let .invalidRemoteURL(url): return "Invalid remote URL: \(url)"
+    case .failedToFindDriver:
+      return """
+      Failed to find WebDriver executable or remote URL to a running driver process.
+      Please make sure that you satisfied one of the followings (smaller item number lower priority):
```
MaxDesiatov, line 39: nit: I'd rephrase it this way:
```suggestion
      Please make sure that you are satisfied with one of the followings (in order fo priority):
```

Sources/CartonCLI/Commands/TestRunners/BrowserTestRunner.swift
```diff
@@ -12,39 +12,165 @@
...
   let testFilePath: AbsolutePath
   let host: String
   let port: Int
+  let headless: Bool
   let manifest: Manifest
   let terminal: InteractiveWriter
+  let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+  let httpClient: HTTPClient
+
+  internal init(
```
MaxDesiatov, line 58: nit: the convention is not to use `internal` as it's the implicit default
```suggestion
  init(
```

Sources/WebDriverClient/WebDriverClient.swift
```diff
@@ -0,0 +1,109 @@
...
+    return url.absoluteString
+  }
+
+  private static func makeRequestBody<R: Encodable>(_ body: R) throws -> HTTPClient.Body {
+    let encoder = JSONEncoder()
+    return try HTTPClient.Body.data(encoder.encode(body))
+  }
+
+  public func goto(url: String) async throws {
+    struct Response: Decodable {}
```
MaxDesiatov, line 95: this inner `Response` struct doesn't seem to be used anywhere?

kateinoigakukun: Good catch

MaxDesiatov (approved):
Great work, thanks! 👏

PR #369, opened 2022-07-16
Toolchain changes for 5.7

yonihemi:
Currently carton itself builds with (host)Swift 5.6, can build SwiftWasm projects with latest 5.7 snapshot.
Still some work needed to make carton build with (host)Swift 5.7.

### DIFF
--- a/Sources/CartonCLI/Commands/TestRunners/BrowserTestRunner.swift
+++ b/Sources/CartonCLI/Commands/TestRunners/BrowserTestRunner.swift
@@ -134,7 +134,7 @@ struct BrowserTestRunner: TestRunner {
   }
 
   func run() async throws {
-    defer { try? httpClient.syncShutdown() }
+    defer { try httpClient.syncShutdown() }
     try Constants.entrypoint.check(on: localFileSystem, terminal)
     let server = try await Server(
       .init(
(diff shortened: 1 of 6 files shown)


Sources/CartonCLI/Commands/TestRunners/BrowserTestRunner.swift
```diff
@@ -134,7 +134,7 @@ struct BrowserTestRunner: TestRunner {
   }
 
   func run() async throws {
-    defer { try httpClient.syncShutdown() }
+    defer { try? httpClient.syncShutdown() }
```
MaxDesiatov, line 137: Is there a specific reason for this change? If so, could you add a code comment explaining why `try?` had to be used?

MaxDesiatov (commented):
Overall seems legit, I only have a question about use of `try?`

Sources/CartonCLI/Commands/TestRunners/BrowserTestRunner.swift
yonihemi: Without this change Xcode 13.4/CLI Swift 5.6.1 build fails with `Call can throw, but errors cannot be thrown out of a defer body`.
I do see that previous PR pass the build with same Xcode so can't really explain it. Don't mind reverting this and keeping this change local for now.

kateinoigakukun (approved):

MaxDesiatov (approved):
Great work, thanks!

PR #370, opened 2022-07-18
Exit non-zero status when test fail on browser

kateinoigakukun:


MaxDesiatov (approved):
Seems legit, great catch! 👏

PR #378, opened 2022-09-14
Use @main attributes

noppefoxwolf:
It can be used the @main attribute if the filename is App.swift. like SwiftUI.
Is there a reason that filename is main.swift?


kateinoigakukun:
Hmm, swiftlint action somehow failed to fetch `main` branch and shows violation diagnostics for unchanged lines. Please ignore it for now.

Also `IntegrationTests.testTokamakBundle` is failing. `carton init` invokes `swift package init`, then rewrites some sources, and `swift package init` creates `main.swift`. So we need to remove it to avoid having both App.swift and main.swift

kateinoigakukun (approved):
Thanks!

PR #381, opened 2022-09-21
Update SwiftPM library version to 5.7

kateinoigakukun:
This allows building packages requiring tool version 5.7.


kateinoigakukun:
@yonihemi Oh, I didn't remember that PR (even though it's just 2 months ago :joy: ) So we can just pick your commits, right?

yonihemi (approved):

PR #384, opened 2022-10-01
Fix SwiftPM dependency warning for template project

gibachan:
In Package.swift of template project, its dependency is declared with [package(name:url:from:)](https://developer.apple.com/documentation/packagedescription/package/dependency/package(name:url:from:)) which has been already deprecated so that Xcode show warining.
<image src="https://user-images.githubusercontent.com/7476703/193413236-5325e91b-f5a2-42d7-b525-802933574e21.png" width="300" />

This PR replace the method with [package(url:from:)](https://developer.apple.com/documentation/packagedescription/package/dependency/package(url:from:)) to fix the warning.

### DIFF
--- a/Sources/CartonKit/Model/Template.swift
+++ b/Sources/CartonKit/Model/Template.swift
@@ -42,7 +42,6 @@ enum TemplateError: Error {
 }
 
 struct PackageDependency: CustomStringConvertible {
-  let name: String
   let url: String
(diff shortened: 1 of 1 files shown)


yonihemi (approved):

yonihemi:
Thanks for the contribution @gibachan !

PR #385, opened 2022-10-07
Allow to save .swift-version even .swift-version is missing

gibachan:
When we have a project which does not contain .swift-version, `carton sdk local wasm-5.7.1-RELEASE` just claims that .swift-version does not exist, even though it is a valid toolchain version. I think it is inconvenient.

This PR makes to allow saving the local toolchain version even if the .swift-version is not present.


kateinoigakukun (approved):

kateinoigakukun:
Thanks :)

PR #395, opened 2023-05-03
Update SwiftPM branch to 5.8 and fix build errors

kateinoigakukun:
- Update SwiftPM to release/5.8 
- Apply swift-format 5.8


Sources/CartonKit/Model/Template.swift
```diff
@@ -92,33 +92,33 @@ extension Template {
   ) throws {
     try fileSystem.writeFileContents(project.path.appending(component: "Package.swift")) {
       var content = """
-      // swift-tools-version:5.6
-      import PackageDescription
-      let package = Package(
-          name: "\(project.name)",\n
-      """
+        // swift-tools-version:5.6
```
yonihemi, line 95:
```suggestion
        // swift-tools-version:5.8
```
We could update the tools version for new projects.

yonihemi (approved):

Sources/CartonKit/Model/Template.swift
kateinoigakukun: I think it should be changed after we updated the default toolchain version

PR #408, opened 2023-09-15
Add linux install script

shial4:
**1. Add linux install  bash script
2. Update readme**  

This Bash script automates the installation and update process for the "Carton" tool. 

1. Defines the destination directory for the Carton binary.
2. Checks if the 'swift' command is available, and if not, installs Swift.
3. Clones the Carton repository from GitHub.
4. Determines the latest release tag and checks out that release.
5. Installs necessary dependencies for Carton on an Ubuntu system.
6. Builds the Carton tool in release mode.
7. If the build is successful, it updates the Carton binary in the specified destination directory.
8. Cleans up temporary files and the cloned repository, regardless of success or failure.

Overall, this script ensures that Carton is installed or updated to the latest release on your system, making it ready for use in Swift WebAssembly development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ To download carton and install it, run the following in your terminal, then foll
 Overall, this script ensures that Carton is installed or updated to the latest release on your system, making it ready for use.
 
 ```
-curl -sSL https://example.com/path/to/your/carton-install.sh | bash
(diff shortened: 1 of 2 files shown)


kateinoigakukun:
Because carton is going to be a SwiftPM Plugin, custom CLI installation is no longer needed. Sorry for leaving it for a long time, and thank you for trying to solve the issue :pray:

